PR #173, opened 2022-01-28
MGDCTRS-78: migrates to java operator sdk v2

rinaldodev:
This migrates to the new Java Operator SDK, version 2.0.2, and also the
Quarkus Operator SDK, to version 3.0.2. For the controller and its
reconcile methods, it's mostly name changes. For the EventSources there
has been a moderate change to the API. For e.g., Events are now
binded to a single ResourceID (before a resource selector). The new
API suggests the use of Informers for EventSources, which would make
for a even simpler implementation, but Informers have internal cache
which could represent a high memory usage in scenarios with many
connectors deployed, that's why we are still using our own Watch-based
EventSources. Existing EventSources had to be changed to trigger
multiple events because of the new 1-1 resource-event relation.


rinaldodev:
- It is possible to try to move our WatcherEventSource to the Java Operator SDK itself, as it is basically a new type of EventSource extra to the SDK, which already offers a few other types.
- More unit and integration tests should be created to test the EventSources.

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/operand/OperandResourceWatcher.java
```diff
@@ -49,6 +55,11 @@ public OperandResourceWatcher(
 
     }
 
+    @Override
+    public String name() {
```
lburgazzoli, line 59: is this used anywhere ?

lburgazzoli:
LGTM, only a small question

lburgazzoli (approved):

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorSecretEventSource.java
```diff
@@ -47,7 +52,11 @@ protected void onEventReceived(Action action, Secret resource) {
             resource.getMetadata().getName(),
             resource.getMetadata().getResourceVersion());
 
-        getEventHandler().handleEvent(
-            new ConnectorSecretEvent(this, resource));
+        List<OwnerReference> ownerReferences = resource.getMetadata().getOwnerReferences();
+        if (!ownerReferences.isEmpty()) {
+            String connectorName = ownerReferences.get(0).getName();
+            getEventHandler().handleEvent(new Event(new ResourceID(connectorName, namespace)));
```
lburgazzoli, line 58: I think we can use the namespace from the owner reference assuming that this kind of secrets are always on the same namespace as the related managed connector (the owner)

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/operand/OperandResourceWatcher.java
rinaldodev: Yes. The SDK cannot register two EventSources with the same name, and the default implementation returns the class name. Since we register multiple EventSources of this type, we need a different name for each instance.

valdar (approved):

PR #174, opened 2022-01-30
Improve camel main configuration set-up

lburgazzoli:
- feat: report liveness DOWN when the supervise controller fails all the attemps to restart a route
- feat: set camel man configuration propertiss with secret


lburgazzoli:
@rinaldodev I've moved some configuration options such as the route controller and exchange pooling to the secret so we configure camel main in a single place

valdar (approved):

PR #198, opened 2022-02-08
Add set up for staging data plane and fix connectors lookup

lburgazzoli:
- feat: add rhoc-staging kustomizations
- fix: lookup connectors in the right namespace


lburgazzoli:
> LGTM except that GitHub complains about missing newlines at the end of some files ;)

will fix it later

valdar (approved):

PR #206, opened 2022-02-11
MGDCTRS-353: Update cos-fleetshard model to the latest fleet manager …

valdar:
…APIs


lburgazzoli (approved):

lburgazzoli:
LGTM

However I think we need to discuss if it is wort to have to maintain 3 openapi templates to generate 3 models and a client with 3 methods. Given that the APIs are not changing very often, I wonder if ti would be simpler to maintain the model directly.

PR #211, opened 2022-02-15
Updated connector json example to take in account new api refactor.

valdar:


etc/examples/connector.json
```diff
@@ -1,15 +1,17 @@
 {
     "kind": "Connector",
     "metadata": {
-        "name": "cos_logger",
-        "kafka_id": ""
+        "name": "cos_logger"
```
lburgazzoli, line 4: is this correct ? I don't think the request has metadata anymore

etc/examples/connector.json
```diff
@@ -1,15 +1,17 @@
 {
     "kind": "Connector",
```
lburgazzoli, line 2: I this has gone

lburgazzoli (changes requested):

etc/examples/connector.json
valdar: done

lburgazzoli (approved):

PR #234, opened 2022-03-04
MGDCTRS-475 Add `data_shape` handling to support "JSON without schema", "JSON" (with Schema Registry using Apicurio) and AVRO (using Apicurio)

rk3rn3r:
* add optional types mapping for "data_shape"
* remove "Properties.class" type handling as it won't occur for connectorSpec
* fix OpenAPI example for `schema_registry.url` (will update OpenAPI spec in upstream fleetmanager repo later)

**open: add integration tests**

https://issues.redhat.com/browse/MGDCTRS-475

### DIFF
--- a/cos-fleet-manager-api/src/openapi/specs/connector_mgmt.yaml
+++ b/cos-fleet-manager-api/src/openapi/specs/connector_mgmt.yaml
@@ -1286,12 +1286,12 @@ components:
           id: "9bsv0s6brfr002pfnkh0"
           client_id: "srvc-acct-162ef2d8-0209-4117-8462-df63c2025c26"
           client_secret: "b144d991-08eb-4472-8ec1-eb034c69231b"
-          url: "foo-9bsv0s6brfr002pfnkh0.kas.acme.com:443"
+          url: "https://bu98.serviceregistry.rhcloud.com/t/51eba005-daft-punk-afe1-b2178bcb523d/apis/registry/v2"
         schema_registry:
           id: "9bsv0s0k8lng031se9q0"
           client_id: "srvc-acct-e025265e-a88a-44fb-8450-ed010eb6bc5c"
           client_secret: "dc5a8298-bed7-422e-8947-f6fda69292ff"
-          url: "bar-9bsv0s2mfca002t9q7sg.srs.acme.com:443"
+          url: "https://bu98.serviceregistry.rhcloud.com/t/51eba005-daft-punk-afe1-b2178bcb523d/apis/registry/v2"
         channel: stable
         connector_type_id: "log_sink_0.1"
         connector:
--- a/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/AbstractApicurioConverter.java
+++ b/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/AbstractApicurioConverter.java
@@ -1,0 +1,36 @@
+package org.bf2.cos.fleetshard.operator.debezium.model;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import org.bf2.cos.fleetshard.api.ManagedConnector;
+import org.bf2.cos.fleetshard.api.SchemaRegistrySpec;
+import org.bf2.cos.fleetshard.api.ServiceAccountSpec;
+import org.bf2.cos.fleetshard.operator.debezium.DebeziumConstants;
+
+public abstract class AbstractApicurioConverter implements KafkaConnectConverter {
+
+    public static final String APICURIO_AUTH_SERVICE_URL = "https://identity.api.openshift.com/auth";
+
+    @Override
+    public Map<String, String> getAdditionalConfig(ManagedConnector config, ServiceAccountSpec serviceAccountSpec) {
+        Map<String, String> additionalConfig = new HashMap<>();
+        additionalConfig.put("apicurio.auth.service.url", APICURIO_AUTH_SERVICE_URL);
+        additionalConfig.put("apicurio.auth.realm", "rhoas");
+
+        SchemaRegistrySpec schemaRegistrySpec = config.getSpec().getDeployment().getSchemaRegistry();
+        if (null == schemaRegistrySpec || null == schemaRegistrySpec.getUrl() || schemaRegistrySpec.getUrl().isBlank()) {
+            throw new RuntimeException("Can't create a schema-based connector without providing a valid 'schema_registry'");
+        }
+        String schemaRegistryURL = schemaRegistrySpec.getUrl();
+
+        additionalConfig.put("apicurio.registry.url", schemaRegistryURL);
+        additionalConfig.put("apicurio.auth.client.id", serviceAccountSpec.getClientId());
+        additionalConfig.put("apicurio.auth.client.secret",
+            "${dir:/opt/kafka/external-configuration/" + DebeziumConstants.EXTERNAL_CONFIG_DIRECTORY + ":"
+                + DebeziumConstants.KAFKA_CLIENT_SECRET_KEY + "}");
+        additionalConfig.put("apicurio.registry.auto-register", "true");
+        additionalConfig.put("apicurio.registry.find-latest", "true");
+        return additionalConfig;
+    }
+}
--- a/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/DebeziumDataShape.java
+++ b/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/DebeziumDataShape.java
@@ -1,0 +1,54 @@
+package org.bf2.cos.fleetshard.operator.debezium.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class DebeziumDataShape {
+
+    private String keyConverterString;
+    private String valueConverterString;
+    private KafkaConnectConverter keyConverter;
+    private KafkaConnectConverter valueConverter;
+
+    private static KafkaConnectConverter createConverter(String converterType) {
+        switch (converterType) {
+            case "JSON":
+                return new ApicurioJsonConverter();
+            case "AVRO":
+                return new ApicurioAvroConverter();
+            default:
+            case "JSON_WITHOUT_SCHEMA":
+            case "JSON without schema":
+                return new KafkaConnectJsonConverter();
+        }
+    }
+
+    @JsonProperty("key")
+    public String getKeyConverterAsString() {
+        return keyConverterString;
+    }
+
+    @JsonProperty("key")
+    public void setKeyConverter(String keyConverterString) {
+        this.keyConverterString = keyConverterString;
+        this.keyConverter = createConverter(keyConverterString);
+    }
+
+    public KafkaConnectConverter getKeyConverter() {
+        return keyConverter;
+    }
+
+    @JsonProperty("value")
+    public String getValueConverterAsString() {
+        return valueConverterString;
+    }
+
+    @JsonProperty("value")
+    public void setValueConverter(String valueConverterString) {
+        this.valueConverterString = valueConverterString;
+        this.valueConverter = createConverter(valueConverterString);
+    }
+
+    public KafkaConnectConverter getValueConverter() {
+        return valueConverter;
+    }
+}
--- a/cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
+++ b/cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
@@ -1,36 +1,32 @@
 package org.bf2.cos.fleetshard.operator.connector;
 
-import java.io.IOException;
-import java.io.StringReader;
-import java.util.Properties;
 import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
-import io.fabric8.kubernetes.client.utils.Serialization;
-
-public class ConnectorConfiguration<S> {
+public class ConnectorConfiguration<S, D> {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConnectorConfiguration.class);
 
-    private final static String PROPERTY_DATA_SHAPE = "data_shape";
-    private final static String PROPERTY_PROCESSORS = "processors";
-    private final static String PROPERTY_ERROR_HANDLER = "error_handler";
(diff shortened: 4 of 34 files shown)


cos-fleet-manager-api/src/openapi/specs/connector_mgmt.yaml
```diff
@@ -1286,12 +1286,12 @@ components:
           id: "9bsv0s6brfr002pfnkh0"
           client_id: "srvc-acct-162ef2d8-0209-4117-8462-df63c2025c26"
           client_secret: "b144d991-08eb-4472-8ec1-eb034c69231b"
-          url: "foo-9bsv0s6brfr002pfnkh0.kas.acme.com:443"
+          url: "https://bu98.serviceregistry.rhcloud.com/t/51eba005-daft-punk-afe1-b2178bcb523d/apis/registry/v2"
```
lburgazzoli, line 1289: this needs to be updated in the fleet manager side, here the openapi definition is only used for generating the code.

rk3rn3r: I know. But I need to do this with a separate PR. Some more things might diverge and I will have a look next week.

rk3rn3r:
I'll add integration tests next week. I'll keep this a draft until this is there. But I think I could kick-off a first review round.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/AbstractApicurioConverter.java
```diff
@@ -0,0 +1,33 @@
...
+
+        SchemaRegistrySpec schemaRegistrySpec = config.getSpec().getDeployment().getSchemaRegistry();
+        if (null == schemaRegistrySpec || null == schemaRegistrySpec.getUrl() || schemaRegistrySpec.getUrl().isBlank()) {
+            throw new RuntimeException("Can't create a schema-based connector without providing a valid 'schema_registry'");
+        }
+        String schemaRegistryURL = schemaRegistrySpec.getUrl();
+
+        additionalConfig.put("apicurio.registry.url", schemaRegistryURL);
+        additionalConfig.put("apicurio.auth.client.id", serviceAccountSpec.getClientId());
+        additionalConfig.put("apicurio.auth.client.secret", serviceAccountSpec.getClientSecret());
```
lburgazzoli, line 28: I think the client secret should be loaded using a file protocol or something like https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/b6402e0b5b62a5eee02e83cde41328794d490e2e/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandSupport.java#L96-L101

rk3rn3r: @lburgazzoli  are you sure? This goes to the Kafka Connect (not connector!) config, which should be invisible to the user. External config should maybe go to objects that needs to go to the connector config?

cos-fleet-manager-api/src/openapi/specs/connector_mgmt.yaml
rk3rn3r: Do we have an idea how we will receive SR URLs? Will it be a full URL like this?

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/AbstractApicurioConverter.java
```diff
@@ -0,0 +1,36 @@
...
+
+public abstract class AbstractApicurioConverter implements KafkaConnectConverter {
+
+    public static final String APICURIO_AUTH_SERVICE_URL = "https://identity.api.openshift.com/auth";
+
+    @Override
+    public Map<String, String> getAdditionalConfig(ManagedConnector config, ServiceAccountSpec serviceAccountSpec) {
+        Map<String, String> additionalConfig = new HashMap<>();
+        additionalConfig.put("apicurio.auth.service.url", APICURIO_AUTH_SERVICE_URL);
+        additionalConfig.put("apicurio.auth.realm", "rhoas");
```
lburgazzoli, line 19: Are those property needed ?  
I recall I did a demo with camel and they were not required

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/DebeziumDataShape.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private static KafkaConnectConverter createConverter(String converterType) {
+        switch (converterType) {
+            case "JSON":
+                return new ApicurioJsonConverter();
+            case "AVRO":
+                return new ApicurioAvroConverter();
+            default:
+            case "JSON_WITHOUT_SCHEMA":
+            case "JSON without schema":
```
lburgazzoli, line 20: do we need those two additional case here ? they don't hurt but at the same time they are not really meaningful as the same case is covered by the default.

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
```diff
@@ -42,7 +42,7 @@ public ConnectorConfiguration(ObjectNode connectorSpec, Class<S> type) {
             LOGGER.error("Missing `data_shape` in connector spec!");
             this.dataShapeSpec = null;
         } else {
-            this.dataShapeSpec = dataShape.deepCopy();
+            this.dataShapeSpec = Serialization.unmarshal(dataShape.deepCopy().toString(), dataShapeType);
```
lburgazzoli, line 45: wonder if a deep copy is really needed

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
```diff
@@ -64,9 +64,9 @@ public ConnectorConfiguration(ObjectNode connectorSpec, Class<S> type) {
...
         connectorSpec.remove(RESERVED_PROPERTIES);
 
-        if (type.isAssignableFrom(String.class)) {
+        if (connectorSpecType.isAssignableFrom(String.class)) {
             this.connectorSpec = (S) connectorSpec.toString();
-        } else if (type.isAssignableFrom(Properties.class)) {
+        } else if (connectorSpecType.isAssignableFrom(Properties.class)) {
             Properties result = new Properties();
             try {
                 result.load(new StringReader(connectorSpec.toString()));
```
lburgazzoli, line 72: is this even supposed to work given that as connector Spec is an ObjectNode

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
```diff
@@ -75,15 +75,15 @@ public ConnectorConfiguration(ObjectNode connectorSpec, Class<S> type) {
             }
             this.connectorSpec = (S) result;
         } else {
-            this.connectorSpec = Serialization.unmarshal(connectorSpec.toString(), type);
+            this.connectorSpec = Serialization.unmarshal(connectorSpec.toString(), connectorSpecType);
```
lburgazzoli, line 78: I think this `connectorSpec.toString()` can be avoided, given that is is already an ObjectNode it should be possible to use the object mapper to convert it to the target type without the need to go through a string before.

cos-fleet-manager-api/src/openapi/specs/connector_mgmt.yaml
rk3rn3r: I see. Okay, this is a customized URL from the RHOSR UI. Then this example should be good!

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/AbstractApicurioConverter.java
rk3rn3r: Was your demo using OAUTH with RHOSR using the Kafka Connect converters? I think this is needed for them, yes!

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/DebeziumDataShape.java
rk3rn3r: Yes. This is only temporarily a default as long as we don't have schema registry support. But the `JSON_WITHOUT_SCHEMA` option will be removed after the next catalog release (where `JSON_WITHOUT_SCHEMA` will change to `JSON without schema`. I think once everything is settled I'll maybe make this an exception eventually?

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
rk3rn3r: I think it isn't needed anymore when we deserialize, true! I will upate this part!

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/AbstractApicurioConverter.java
lburgazzoli: I was using the serdes from apicurio
If they are needed, IMHO, hey should not be hard-coded  but provided by a configuration option

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/DebeziumDataShape.java
lburgazzoli: maybe add a comment to explain why we have two explicit cases that led to a default behavior, you know, for the next one that have to understand this case.

also we need to clarify how we support schemes with PM, maybe the registry will be needed anyway when it is about dealing with schemes.

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
rk3rn3r: That is code that was previously used for the connectorSpec type in `org.bf2.cos.fleetshard.support.resources.Secrets#extract(..)`. I just copied it here as Deserialization for this moved here for the properties that were left. Not sure if we need it. I just copied it with the previous-previous PR.

lburgazzoli: because in a secret you may have a string used to represent an application.properties  i.e.

```yaml
data:
    application.properties: ${base64}
```

So that method was an helper to extract properties from a secret data entry

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
rk3rn3r: Agree! We don't have to de-serialize YAML here (which is covered by `Serialization.unmarshal` too)

rk3rn3r: This applies to `dataShapeSpec` above too.

rk3rn3r:
@lburgazzoli I applied your feedback from the resolved conversations above.
I don't know what was the reason behind `Properties.class` part I think I saw some code that could provide some spec as `Properties` instance. Dunno if it applies to the connectorSpec too? Maybe you can answer on the conversation above.
If you agree to my other comment that PR might be ready to be merged? wdyt?

lburgazzoli:
> @lburgazzoli I applied your feedback from the resolved conversations above. I don't know what was the reason behind `Properties.class` part I think I saw some code that could provide some spec as `Properties` instance. Dunno if it applies to the connectorSpec too? Maybe you can answer on the conversation above. If you agree to my other comment that PR might be ready to be merged? wdyt?

I think the option to unwrap a secret key as a properties is needed for some integration tests but it does not seem to make sense in the `ConnectorConfiguration` class.

The other point that we should clear is if the `apicurio.auth.service|realm` is really needed or not as this may have some impacts i.e. we may need to provide such info as a secret/configmap instead of hard-coding it.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/DebeziumDataShape.java
rk3rn3r: @lburgazzoli will revisit this when the catalog is changed. Then we should also have a decision that I put on the list for today's backlog refinement session.

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorConfiguration.java
rk3rn3r: ah. so this is not the case for connectorSpec, true! I will drop the code then

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/model/AbstractApicurioConverter.java
rk3rn3r: wdym with "provided by a configuration option? They are conditional properties that gets added for either the key or value converter dependent of the selected `data_shape`. So this is the case!

lburgazzoli: I mean that the values from `apicurio.auth.service.url` and the `apicurio.auth.realm` should not be hard-coded but they should be retrieved from a secret.

rk3rn3r: Thx @lburgazzoli! Now I gotcha!
I filed MGDCTRS-584](https://issues.redhat.com/browse/MGDCTRS-584) until we know values for different environments and it involves deployment of such secrets too.

rk3rn3r: Can we add this later because this PR waits a while now and @ani-sha needs this for her work because it enhanced the cucumber tests that she needs for her work on the pull secret and we need the pull secret asap!

lburgazzoli (approved):

PR #240, opened 2022-03-15
MGDCTRS-257: Explicit set properties in KameletBinding

lburgazzoli:


oscerd (approved):

valdar:
The counter argument to all of this is that if a kamelet property is missconfigured there might be sensitive data spilled around...

valdar (approved):

rinaldodev (approved):

PR #245, opened 2022-03-16
feat(MGDCTRS-614): prevent connector from being ready before klb deploy

rinaldodev:
Waits for the Ready condition on the KameletBinding to be True before reporting the connector as Ready. 
Previously only the KameletBinding phased was being considered.


oscerd (approved):

lburgazzoli (approved):

oscerd:
It seems it's failing on some repository, I can confirm it's the same for camel-kamelets upstream, while it is working for Camel.

rinaldodev:
Thanks for the info. Workflow has run successfully on main: https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/actions/runs/1994220210

PR #249, opened 2022-03-17
explicit set permissions for mounted volumes as the service does not run anymore as root user

lburgazzoli:


Naros:
Tested locally and the cos-fleetshard-sync PoD starts successfully :pray:

PR #250, opened 2022-03-18
Pin version for next release

lgarciaaco:


.github/project.yaml
```diff
@@ -1,4 +1,3 @@
-name: Managed Connector Services - Camel Meta Service
+name: Managed Connector Services
 release:
-  current-version: 0.0.1
-  next-version: 0.0.2-SNAPSHOT
\ No newline at end of file
+  current-version: v1.0.3
```
lburgazzoli, line 3: Wonder how this is really end up being used, my understanding is that as today each build  will generates ad image on the rhoas repo with 3 tags:

- latest
- ${SHA}
- ${current-version}

lgarciaaco: That's the plan, to reference the current version in the Addons

lburgazzoli: Ok, we need to be sure this version is bumped before we pin the image everywhere else

PR #252, opened 2022-03-21
MGDCTRS-332 Apply Strimzi selector to Debezium CRs

Naros:
https://issues.redhat.com/browse/MGDCTRS-332

- [x] Applies `STRIMZI_CUSTOM_RESOURCE_SELECTOR` label to Debezium operator resources

### DIFF
--- a/cos-fleetshard-operator-debezium/README.adoc
+++ b/cos-fleetshard-operator-debezium/README.adoc
@@ -29,6 +29,17 @@ export COS_OPERATOR_VERSION==${some_value}
 ./mvnw -Dlocal
 ----
 
+[IMPORTANT]
+====
+The `COS_OPERATOR_ID` should be synchronized with the value assigned by the addon bundle.
+As of this commit, this would be `cos-fleetshard-operator-debezium-1.0.1`.
+
+Please check the bundles repository for this operator's latest cluster service version configuration and make sure this is set to the same value as the container's `COS_OPERATOR_ID` environment variable.
+
+Additionally, the `COS_OPERATOR_VERSION` should be synchronized from the same location but should match the cluster service version configuration's environment variable of the same name.
+As of this commit, this would be `1.5.0`.
+====
+
 By default, the application searches for:
 
 * **ConfigMap**: cos-fleetshard-operator-debezium-config
--- a/etc/kubernetes/operator-debezium/lgarciaac/kustomization.yaml
+++ b/etc/kubernetes/operator-debezium/lgarciaac/kustomization.yaml
@@ -23,7 +23,7 @@ patchesJson6902:
         path: /spec/template/spec/containers/0/env/-
         value:
           name: COS_OPERATOR_ID
-          value: cos-fleetshard-operator-debezium
+          value: "cos-fleetshard-operator-debezium-1.0.1"
       - op: add
         path: /spec/template/spec/containers/0/env/-
         value:
--- a/etc/kubernetes/operator-debezium/local/kustomization.yaml
+++ b/etc/kubernetes/operator-debezium/local/kustomization.yaml
@@ -25,7 +25,7 @@ patchesJson6902:
         path: /spec/template/spec/containers/0/env/-
         value:
           name: COS_OPERATOR_ID
-          value: cos-fleetshard-operator-debezium
+          value: "cos-fleetshard-operator-debezium-1.0.1"
       - op: add
         path: /spec/template/spec/containers/0/env/-
         value:
--- a/etc/kubernetes/operator-debezium/rh-fuse/kustomization.yaml
+++ b/etc/kubernetes/operator-debezium/rh-fuse/kustomization.yaml
@@ -23,7 +23,7 @@ patchesJson6902:
         path: /spec/template/spec/containers/0/env/-
         value:
           name: COS_OPERATOR_ID
-          value: cos-fleetshard-operator-debezium
+          value: "cos-fleetshard-operator-debezium-1.0.1"
       - op: add
         path: /spec/template/spec/containers/0/env/-
         value:
--- a/etc/kubernetes/operator-debezium/rhoc-staging/kustomization.yaml
+++ b/etc/kubernetes/operator-debezium/rhoc-staging/kustomization.yaml
@@ -23,7 +23,7 @@ patchesJson6902:
         path: /spec/template/spec/containers/0/env/-
         value:
           name: COS_OPERATOR_ID
-          value: cos-fleetshard-operator-debezium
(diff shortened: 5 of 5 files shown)


cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
```diff
@@ -217,4 +224,10 @@ public boolean delete(ManagedConnector connector) {
 
         return kctr && kc && secret;
     }
+
+    private String getOperatorName(ManagedConnector connector) {
+        return connector.getStatus().getConnectorStatus().getAssignedOperator().getId() +
+            "-" +
+            connector.getStatus().getConnectorStatus().getAssignedOperator().getVersion();
```
lburgazzoli, line 231: is this correct ? I recall the in latest discussion I had with @valdar, the operator id would be set to a unique value in the deployment as an env var.

Naros: Hi @lburgazzoli, this works locally but this is likely due to what is set as the default values in the `kustomization.yaml` inside of the `/etc/kubernetes/operator-<operator>/local/` directories.  Let me know if this should be changed to something else.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
```diff
@@ -150,6 +156,7 @@ public DebeziumOperandController(KubernetesClient kubernetesClient, DebeziumOper
             .withMetadata(new ObjectMetaBuilder()
                 .withName(connector.getMetadata().getName())
                 .addToLabels(STRIMZI_DOMAIN + "cluster", connector.getMetadata().getName())
+                .addToLabels(BF2_DOMAIN + "connector.operator", operatorName)
```
rk3rn3r, line 159: Looks like what we really use here is a label name (`BF_DOMAIN + connector.operator` in all cases. Maybe we can just use a single constant string instead of appending it multiple times?

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumConstants.java
```diff
@@ -24,6 +24,8 @@
 
     public static final LocalObjectReference IMAGE_PULL_SECRET = new LocalObjectReference("addon-pullsecret");
 
+    public static final String BF2_DOMAIN = "cos.bf2.org/";
```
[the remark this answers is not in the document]
Naros: I can certainly do that, I think the much more important point is the discussion above with Luca about the actual values that are being applied via the env vars and whether that has any direct/critical impact on this change.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
rk3rn3r: Maybe we can make those configurable from `application.properties`?

lburgazzoli: @Naros yes they need to be changed in the repo, likely the operator id need to be changed i.e. to something unique (looks at the addon/bundle)
@rk3rn3r it is already the case, the id and version are from an env var/property

rk3rn3r: @lburgazzoli yes, but this code makes the assumption that there is this pattern to generate the label from the operator id and version. A config option that just sets the full label name is more flexible and avoids code changes when the pattern changes.

lburgazzoli: That label should just use the operator id, it should not concatenate things (see my first comment)

rk3rn3r: Gotcha now! Sorry!

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
```diff
@@ -136,10 +137,15 @@ public DebeziumOperandController(KubernetesClient kubernetesClient, DebeziumOper
...
 
+        final String operatorName = getOperatorName(connector);
+        LOGGER.debug("Assigned {}connector.operator as {} for connector {}", BF2_DOMAIN, operatorName,
+            connector.getMetadata().getName());
+
         final KafkaConnect kc = new KafkaConnectBuilder()
             .withApiVersion(Constants.RESOURCE_GROUP_NAME + "/" + KafkaConnect.CONSUMED_VERSION)
             .withMetadata(new ObjectMetaBuilder()
                 .withName(connector.getMetadata().getName())
+                .addToLabels(BF2_DOMAIN + "connector.operator", operatorName)
```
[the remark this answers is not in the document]
Naros: So there is still something that isn't clear to be @valdar and @lburgazzoli.

If I adjust the `kustomization.yaml` files under `operator-debezium/local` and set the `COS_OPERATOR_ID` as it is from the addon to be `"cos-fleetshard-operator-debezium-1.0.1"`, when a connector create request comes into the operator it deems it as not assigned because the operator id in the pod is the `COS_OPERATOR_ID` but the assigned check reports this:
```
2022-03-22 16:40:50,882 DEBUG [org.bf2.cos.fleetshard.operator.connector.ConnectorController] (EventHandler-connector) Connector default/mctr-c8svnsu3bd8ifa234ni0 is not managed by this operator (assigned=cos-fleetshard-operator-debezium, operating=null).
```

So am I still just misunderstanding what needs to be adjusted to test this locally?

lburgazzoli (approved):

lburgazzoli:
@Naros maybe add a note here: https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/tree/main/cos-fleetshard-operator-debezium#local-profile to tell people what is the right value to put on the two env vars

lburgazzoli (approved):

PR #253, opened 2022-03-22
Add operator bundles and addon test workflow

asmigala:


lburgazzoli:
@asmigala is this auto-generated ?

asmigala:
@lburgazzoli no, this is copied from the internal managed-tenants-bundle repo. We discussed this with @lgarciaaco , that currently the bundles are only downstream in the managed tenants repo and it would be nice to have them upstream as well, with the ability to test them before building the addon and releasing to stage.

lburgazzoli:
fine with me, I hoped you were able to auto generate them :)

heiko-braun:
Managed-tenants downstream is actually a two step process: a) build bundles and index image b) PR to rollout the new bundle. The second step could be used to take the bundles and test them if the copy here becomes unwieldly

lgarciaaco:
I honestly like the idea of having full control over the testing workflows. We can run the harness test midstream nevertheless.

Secondly, we work upstream, once we are done we bring changes midstream where builds take place and deployments run. Working directly agains midstream would be new.

lburgazzoli:
@heiko-braun @lgarciaaco what should we do with this one ?

heiko-braun:
@asmigala @lgarciaaco and I agreed to take another approach and rely on the managed tenants repo as the source of truth. We have ways to test the bundles produced over there, albeit manually for now. We will raise an RFE to introduce UMB messaging for QE to hook up test runs in the future

PR #255, opened 2022-03-23
chore: correct pull secret creation instruction

asmigala:


README.adoc
```diff
@@ -92,7 +92,7 @@ In order to use private image on quay a pull secret need to be crated:
 [source,shell]
 ----
 kubectl create secret generic addon-pullsecret \
-    --from-file=${path of rhoas-pull-docker} \
+    --from-file=.dockerconfigjson=${path of rhoas-pull-docker} \
```
lburgazzoli, line 95: that's ok, but then please add .dockerconfigjson to the gitignore as I'm pretty sure it would be leaked

asmigala: This only sets the key name inside the kubernetes secret - it does not actually create the file locally.

lburgazzoli: oh sorry, I miread it

lburgazzoli (approved):

PR #258, opened 2022-03-24
MGDCTRS-617 Enable and add Prometheus metrics for Debezium connectors

rk3rn3r:
closes https://issues.redhat.com/browse/MGDCTRS-600

### DIFF
--- a/cos-fleetshard-operator-debezium/pom.xml
+++ b/cos-fleetshard-operator-debezium/pom.xml
@@ -12,7 +12,6 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>cos-fleetshard-operator-debezium</artifactId>
-    <packaging>jar</packaging>
 
     <properties>
         <!-- CRD -->
@@ -79,6 +78,14 @@
         </dependency>
 
         <dependency>
+            <groupId>com.github.spotbugs</groupId>
+            <artifactId>spotbugs-annotations</artifactId>
+            <version>${spotbugs.version}</version>
+            <optional>true</optional>
+            <scope>compile</scope>
+        </dependency>
+
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-container-image-jib</artifactId>
         </dependency>
@@ -119,6 +126,13 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.bf2</groupId>
+                        <artifactId>cos-fleetshard-api-deployment</artifactId>
+                        <version>${project.version}</version>
+                    </dependency>
+                </dependencies>
                 <executions>
                     <execution>
                         <id>build</id>
--- a/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
+++ b/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
@@ -1,8 +1,11 @@
 package org.bf2.cos.fleetshard.operator.debezium;
 
+import java.io.File;
+import java.nio.file.Files;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.TreeMap;
 
 import javax.inject.Singleton;
@@ -21,6 +24,9 @@ import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import io.fabric8.kubernetes.api.model.ConfigMap;
+import io.fabric8.kubernetes.api.model.ConfigMapBuilder;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.ObjectMetaBuilder;
 import io.fabric8.kubernetes.api.model.Secret;
@@ -30,6 +36,8 @@ import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.dsl.base.ResourceDefinitionContext;
 import io.strimzi.api.kafka.model.ClientTlsBuilder;
 import io.strimzi.api.kafka.model.Constants;
+import io.strimzi.api.kafka.model.ExternalConfigurationReferenceBuilder;
+import io.strimzi.api.kafka.model.JmxPrometheusExporterMetricsBuilder;
 import io.strimzi.api.kafka.model.KafkaConnect;
 import io.strimzi.api.kafka.model.KafkaConnectBuilder;
 import io.strimzi.api.kafka.model.KafkaConnectSpecBuilder;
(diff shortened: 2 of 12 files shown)


cos-fleetshard-operator-debezium/pom.xml
```diff
@@ -64,6 +64,10 @@
             <groupId>org.bf2</groupId>
             <artifactId>cos-fleetshard-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.bf2</groupId>
+            <artifactId>cos-fleetshard-api-deployment</artifactId>
```
lburgazzoli, line 69: is this required ?

rk3rn3r: Yes. When you do a fresh checkout the dependency seems to be missing and the project doesn't compile.
I think when you delete your local `~/.m2/repository/org/bf2` cache then you can see it stops compiling. A `./mvnw clean package -am -pl $module_name` won't work as it won't build that dependent module.
If you have a better solution I can do that.

rk3rn3r:
@lburgazzoli  any idea why GH Actions don't kick in?

lburgazzoli:
> @lburgazzoli any idea why GH Actions don't kick in?

https://www.githubstatus.com/

rk3rn3r:
Yes, edited above

cos-fleetshard-operator-debezium/pom.xml
lburgazzoli: I think this is an issue with quarkus but the issue here is that for quarkus pov, a `-deployment` dependency should not be part of the runtime

rk3rn3r: @lburgazzoli really? So we call the module `-depl0yment` now, or how we gonna solve it?
On the other hand, it works when I install that module once and have it in my Maven cache.

heiko-braun:
@rk3rn3r Can you add a section to [1] outlining the SLIs you have enabled with this?

[1] https://docs.google.com/document/d/1KY0Uqf_wyZOkUKhuPCBHN6HYIqfosGTvVkq_YXNrOqA/edit

cos-fleetshard-operator-debezium/pom.xml
lburgazzoli: that is how quarkus works: the `-deployment` are used a build time and are not included in the runtime.
the dependency is specified by some metadata included in the `cos-fleetshard-api` module and resolved by the quarkus maven plugin.

see https://quarkus.io/guides/building-my-first-extension

rk3rn3r: You mean the `cos-fleetshard` parent? Yeah, but often I don't want to build Camel when I only change Debezium. And I expect the build to work.
But back to the topic, can we keep it like this, or can I fix that dependency issue in a different way?

lburgazzoli: > But back to the topic, can we keep it like this, or can I fix that dependency issue in a different way?

I would really like to avoid that as this has an impact on the produced artifacts and how quarkus works

rk3rn3r: Agreed. But how? 🤔

lburgazzoli: just remove the additional deps

you may need to rebuild the project instead of individual module till we find a solution :)

rk3rn3r: Ha! Found the solution!

cos-fleetshard-operator-debezium/pom.xml
```diff
@@ -119,6 +119,13 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.bf2</groupId>
+                        <artifactId>cos-fleetshard-api-deployment</artifactId>
+                        <version>${cos-fleetshard-api.version}</version>
+                    </dependency>
+                </dependencies>
```
rk3rn3r, line 128: Like this! As plugin dependency! @lburgazzoli

lburgazzoli: can you do the same on the camel module ?

cos-fleetshard-operator-debezium/pom.xml
```diff
@@ -119,6 +119,13 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.bf2</groupId>
+                        <artifactId>cos-fleetshard-api-deployment</artifactId>
+                        <version>${cos-fleetshard-api.version}</version>
```
lburgazzoli, line 126: can this be ${project,version} ?

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
```diff
@@ -92,10 +116,11 @@ public DebeziumOperandController(KubernetesClient kubernetesClient, DebeziumOper
             .addToData(KAFKA_CLIENT_SECRET_KEY, serviceAccountSpec.getClientSecret())
             .build();
 
-        PodTemplate podTemplate = new PodTemplateBuilder()
-            .withImagePullSecrets(IMAGE_PULL_SECRET)
+        ConfigMap kafkaConnectMetricsConfigMap = new ConfigMapBuilder()
```
lburgazzoli, line 119: This is fine for now but given that the configmap is always the same, can we create a config map once when the operator starts ? 

Eventually such config map can also be provided as part of the addon.

cos-fleetshard-operator-debezium/pom.xml
rk3rn3r: I don't know. This is a property that we already have and use everywhere. Maybe it's a valid use case to progress with the api package code without using the newer version at the same time? This is just how it is specified in the `cos-fleetshard` dependencyManagement section.

lburgazzoli: The `cos-fleetshard-api` is a module of the project and defines the CRD of the operator
I don't see a good use case to keep it on a different version.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
rk3rn3r: See line 71 in the same file: https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/3e9bcc813b4cbe0123b155020d34c98aca40c51b/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java#L71
:wink:

lburgazzoli: ah great, didn't spot it

rk3rn3r: The intention is/was to maybe have some connector specific config eventually. When we see it won't be needed, I agree to put it into the bundle/addon config is the way to go. That's why I put the comment above.

cos-fleetshard-operator-debezium/pom.xml
rk3rn3r: Done

cos-fleetshard-operator-debezium/pom.xml
rk3rn3r: Done already

lburgazzoli (approved):

PR #259, opened 2022-03-24
cos-fleetshard-sync: enable multi tenancy support 

lburgazzoli:
- feat(cos-fleetshard-sync): create the namespaces on the data plane (MGDCTRS-596)
- feat(cos-fleetshard-sync): add a feature flag to enable/disable tenancy support (MGDCTRS-596)


heiko-braun:
Why the feature flag?

lburgazzoli:
to deploy it now without requiring synchronized deployment of the fleet manager

heiko-braun:
How will the feature be enabled? I assume it doesn’t require redeployment?

lburgazzoli:
> How will the feature be enabled? I assume it doesn’t require redeployment?

correct, there is a configuration option to enable/disable tenancy support.

to be 100% clear, right now it works the other way as tenancy is enable by default but we can turn it down in case the manged is not deployed in time

cos-fleetshard-sync/src/main/kubernetes/kubernetes.yml
```diff
@@ -7,20 +7,17 @@ imagePullSecrets:
   - name: addon-pullsecret
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
```
valdar, line 10: namespaces as well should be added at ClusterRole

valdar (changes requested):

cos-fleetshard-sync/src/main/kubernetes/kubernetes.yml
lburgazzoli: done

valdar (approved):

valdar (approved):

lgarciaaco (approved):

PR #263, opened 2022-03-28
Container name should match operator name

lgarciaaco:
I need to be like in stage, where container name match the operator name


lburgazzoli:
@lgarciaaco can you change it also on the mother profiles ?

lburgazzoli (approved):

PR #265, opened 2022-03-28
feat(MGDCTRS-639): makes operators cluster-wide

rinaldodev:
Changes the behavior of debezium and camel operators so they will watch
the entire cluster for changes on ManagedConnector and other resources.

The property cos.connectors.namespace is not used anymore. Resources are
now created in the same namespace where the ManagedConnector resource
already is.

Roles and RoleBindings were changed to cluster.

### DIFF
--- a/etc/kubernetes/operator-camel/kcp/app-config/application.properties
+++ b/etc/kubernetes/operator-camel/kcp/app-config/application.properties
@@ -9,5 +9,4 @@ quarkus.log.category."io.fabric8.kubernetes.client.internal.VersionUsageUtils".l
 # operator
 #
 cos.connectors.target-labels[0] = "kcp.dev/cluster"
-cos.connectors.namespace = ${kubernetes.namespace}
 cos.operators.namespace  = ${kubernetes.namespace}
--- a/etc/kubernetes/operator-camel/lgarciaac/app-config/application.properties
+++ b/etc/kubernetes/operator-camel/lgarciaac/app-config/application.properties
@@ -8,5 +8,4 @@ quarkus.log.category."io.fabric8.kubernetes.client.internal.VersionUsageUtils".l
 #
 # operator
 #
-cos.connectors.namespace = ${kubernetes.namespace}
 cos.operators.namespace   = ${kubernetes.namespace}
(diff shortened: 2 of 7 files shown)


lburgazzoli (changes requested):
The roles & role binding need to be updated to be cluster scoped.

i.e. in https://github.com/rinaldodev/cos-fleetshard-1/blob/feat/cluster-mode-operators/cos-fleetshard-operator-camel/src/main/kubernetes/kubernetes.yml, we define a `cos-fleetshard-operator-camel` role that lists the resources the operator can access using `rbac.authorization.k8s.io/v1:Role` however, that's limits the visibility to the namespace in which the operator runs.

We should migrate:
- Role -> ClusterRole
- RoleBinding -> ClusterRoleBinding

cos-fleetshard-it/cucumber/src/main/java/org/bf2/cos/fleetshard/it/cucumber/ConnectorContext.java
```diff
@@ -52,7 +51,7 @@
     private final List<ManagedConnector> history = new CopyOnWriteArrayList<>();
 
     @Inject
-    private CosFeatureContext cosCtx;
+    CosFeatureContext cosCtx;
```
valdar, line 54: Why the change in scope?

valdar (changes requested):

cos-fleetshard-it/cucumber/src/main/java/org/bf2/cos/fleetshard/it/cucumber/ConnectorContext.java
rinaldodev: Quarkus complains since it has to use reflection to inject it, and from looking at other Injection points we seem to be following this pattern of package-private.

valdar (approved):

heiko-braun:
@rinaldodev I guess this requies a larger change to the managed-tenants addon as well. Do you plan to provide it too?

lburgazzoli (approved):

PR #275, opened 2022-04-01
Restructure k8s manifests to suport bundle generation

lburgazzoli:
- feat: re-structure kubernetes manifest to support bundle generation
    - the kubernetes manifest folder has been restructured and simplified
    - operators and sync configuration has been simplified to reduce the need to apply kustomize patch
    - the deployment to dev cluster has been simplified ans may be replaced by ArgoCD at some point
- feat: remove tenancy feature flag
    - as part of the simplification of the set-up the feature flag to disable multi tenancy has been removed given that it is not more needed to support non multi tenant managers
- chore: update connector example


valdar (approved):
lgtm

etc/examples/argocd-app.yaml
```diff
@@ -0,0 +1,20 @@
...
+kind: Application
+metadata:
+  name: rhoc
+spec:
+  destination:
+    namespace: cos
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    path: etc/kubernetes/manifests/overlays/lburgazzoli
```
valdar, line 11: is this a leftover?

etc/examples/argocd-app.yaml
```diff
@@ -0,0 +1,20 @@
...
+    repoURL: https://github.com/lburgazzoli/bf2-cos-fleetshard
```
valdar, line 12: same here?

valdar (changes requested):

etc/examples/argocd-app.yaml
lburgazzoli: yes, it will be removed before marking the PR as ready

etc/examples/argocd-app.yaml
lburgazzoli: yes, it will be removed before marking the PR as ready

cos-fleetshard-poms/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <parent>
+        <groupId>org.bf2</groupId>
+        <artifactId>cos-fleetshard</artifactId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>cos-fleetshard-poms</artifactId>
```
rk3rn3r, line 14: What's exactly the role of this module?

lburgazzoli: It is ti simplify the poms for the operator/sync as they share almost the same configuration for plugins and set-up for container image build.

valdar (approved):
lgtm

PR #278, opened 2022-04-04
Add missing resource in debezium's ClusterRole

lburgazzoli:
- fix: add missing role to debezium's kubernetes template
- chore: regen


lburgazzoli:
@rk3rn3r for some reason the config map resource was missing from the debezium setup

PR #279, opened 2022-04-04
fix: add missing role to debezium's kubernetes template

lburgazzoli:


cos-fleetshard-operator-debezium/src/main/kubernetes/kubernetes.yml
```diff
@@ -34,6 +34,7 @@ rules:
   - ""
   resources:
   - secrets
+  - configmaps
```
rk3rn3r, line 37: What is correct with or without quotes?

lburgazzoli: in this case it is the same (the other file is generated by k8s client so it depend on the yaml serdes setup)

rk3rn3r (approved):
LGTM

PR #283, opened 2022-04-05
MGDCTRS-727 Fix metrics config to be a UTF-8 String not a raw byte array

rk3rn3r:
closes https://issues.redhat.com/browse/MGDCTRS-727


lburgazzoli (approved):
A smoke test would be a nice addition

PR #321, opened 2022-04-08
feat(sync): delete resources when addon is deleted (MGDCTRS-768)

rinaldodev:
Don't merge yet.


cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/cleanup/AddonConfigMapWatcher.java
```diff
@@ -0,0 +1,91 @@
...
+import org.bf2.cos.fleetshard.support.watch.AbstractWatcher;
+import org.bf2.cos.fleetshard.sync.FleetShardSync;
+import org.bf2.cos.fleetshard.sync.FleetShardSyncConfig;
+import org.bf2.cos.fleetshard.sync.client.FleetShardClient;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+
+public class AddonConfigMapWatcher extends AbstractWatcher {
```
lburgazzoli, line 20: I would put this as part of the resources sync task instead of using a watcher as in case something goes wrong, then there won't be any new configmap creation event  so we may leave leftovers

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/cleanup/AddonConfigMapWatcher.java
```diff
@@ -0,0 +1,91 @@
...
+                    .withLabel(Resources.LABEL_CLUSTER_ID, clusterId)
+                    .delete();
+
+            LOGGER.info("Deleting all Namespaces that belong to cluster {}", clusterId);
+            kubernetesClient.namespaces()
+                    .withLabel(Resources.LABEL_CLUSTER_ID, clusterId)
+                    .delete();
+
+            LOGGER.info("Deleting {} in namespace {} to signal that addon deletion should continue", OPERATOR_KIND, config.namespace());
+            Boolean deleteCsvOp = kubernetesClient.genericKubernetesResources(OPERATORS_API, OPERATOR_KIND)
```
lburgazzoli, line 83: is this really needed ? 
is so there should be a related cluster role

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/cleanup/AddonConfigMapWatcher.java
rinaldodev: I understand we need a way to not have leftovers if something fails, but I don't get how we could do this in resources sync task. We need to execute this as soon as someone tries to delete the addon, and and that is precisely when this ConfigMap is created.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/cleanup/AddonConfigMapWatcher.java
rinaldodev: Yes, this is what signals that we have finished deleting other namespaces and that the addon should continue with the normal deletion process.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/cleanup/AddonConfigMapWatcher.java
rinaldodev: I've made some changes. See if this is more inline with what you meant.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/AddonCleanup.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public void run() {
+        recorder.record(this::cleanup);
+    }
+
+    public void cleanup() {
+        String clusterId = config.cluster().id();
+        LOGGER.info("Deleting all ManagedConnector resources that belong to cluster {}", clusterId);
+        kubernetesClient.resources(ManagedConnector.class)
```
lburgazzoli, line 69: we may not need this one as the connectors are automatically deleted by deleting the ns.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/AddonCleanup.java
```diff
@@ -0,0 +1,88 @@
...
+                .delete();
+
+        LOGGER.info("Deleting all Namespaces that belong to cluster {}", clusterId);
+        kubernetesClient.namespaces()
+                .withLabel(Resources.LABEL_CLUSTER_ID, clusterId)
+                .delete();
+
+        LOGGER.info("Deleting {} in namespace {} to signal that addon deletion should continue", OPERATOR_KIND,
+                config.namespace());
+        kubernetesClient.genericKubernetesResources(OPERATORS_API, OPERATOR_KIND)
```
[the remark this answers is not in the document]
rinaldodev: Added check along with retries and a limit for retry.

lburgazzoli (approved):

PR #328, opened 2022-04-13
MGDCTRS-802 Map `TimeoutException` condition-reason to "The configured Kafka Cluster is unreachable." as error message

rk3rn3r:
closes https://issues.redhat.com/browse/MGDCTRS-802


rk3rn3r:
@lburgazzoli yes, that's basically the idea of mapping this. there should be no other reason that I can guess now. But we need to test this in real life.

rk3rn3r:
Mind having a look @lburgazzoli ?

lburgazzoli (approved):

PR #335, opened 2022-04-20
Sanitize names before applying resources

lburgazzoli:
- MGDCTRS-823: Sanitize names before applying resources
- feat: parallel it

### DIFF
--- a/.github/workflows/build-it.yaml
+++ b/.github/workflows/build-it.yaml
@@ -29,7 +29,7 @@ jobs:
       - name: "Download Maven Repo"
         uses: actions/download-artifact@v2
         with:
-          name: maven-repo
+          name: maven-repo-${{ github.run_id }}
           path: ..
       - name: "Extract Maven Repo"
         shell: bash
--- a/.github/workflows/build-it.yaml
+++ b/.github/workflows/build-it.yaml
@@ -1,0 +1,52 @@
+name: Build IT
+
+on:
+  workflow_call:
+    inputs:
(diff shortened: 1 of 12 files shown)


lburgazzoli:
@rinaldodev mind having a look ?

I've tried to split the PR tests in separate jobs to see if helps speeding up tests but unsure ...

rinaldodev:
> @rinaldodev mind having a look ?
> I've tried to split the PR tests in separate jobs to see if helps speeding up tests but unsure ...

Looks good. The problem I see is that setting up Kind is a huge overhead to IT tests, and we still need to wait the 5min build before even starting to run the tests in parallel anyway. That said, might not be a huge improvement for now (20 to 30% reduction in my super fast analysis), but this will let us add more tests without increasing test time too much.

lburgazzoli:
> > @rinaldodev mind having a look ?
> > I've tried to split the PR tests in separate jobs to see if helps speeding up tests but unsure ...
> 
> Looks good. The problem I see is that setting up Kind is a huge overhead to IT tests, and we still need to wait the 5min build before even starting to run the tests in parallel anyway. That said, might not be a huge improvement for now (20 to 30% reduction in my super fast analysis), but this will let us add more tests without increasing test time too much (which alone is a good improvement IMO).

Yeah I wasn't really unsure about the change but, tried ... we also need to upload/download the maven repo as jobs can't share any folder ...

rinaldodev (approved):

PR #345, opened 2022-04-26
Fix fo MGDCTRS-866, MGDCTRS-865, MGDCTRS-867

lburgazzoli:
- MGDCTRS-867: namespace name sanitation does not work
- MGDCTRS-865: error broadcasting events
- MGDCTRS-866: don't skip namespace creation if there are still connectors deployed on it


lburgazzoli:
the fix for MGDCTRS-866 assumes that the a namespace can be deleted only when 

- status.state == deleting | deleted
- status.connectors_deployed == 0

@dhirajsb can you please validate ?

dhirajsb:
@lburgazzoli since namespace status is set by fleetshard, the status field `status.connectors_deployed` is not set by the fleet manager. It's only used for potential UI display by the fleet manager. 

To verify that there are no connectors deployed, you can query deployments with `search=namespace_id=?`.

dhirajsb:
If the state is `deleted`, there is no need to check for deployments.

lburgazzoli:
> @lburgazzoli since namespace status is set by fleetshard, the status field `status.connectors_deployed` is not set by the fleet manager. It's only used for potential UI display by the fleet manager.
> 
> To verify that there are no connectors deployed, you can query deployments with `search=namespace_id=?`.

Ok but it seems a little bit confusing as:
- the state is set by the fleet manager 
- the connectors_deployed is set by the fleetshard

We should not have resources where the status has multiple owners as it makes very hard to reason about what to do.
The fleetshard should clearly state what the status should be so the fleetshard should see a "deleting" phase only when the fleet shard can really start deleting the namespace.

dhirajsb:
Both clusters and namespace use `deleting` as an interim state to denote that cascade delete has started. That allows fleetmanager to look for empty namespaces and empty clusters later to cleanly cleanup resources. 

Do you need the fleetmanager to set the `connectors_deployed` field when retrieving namespaces? Would that make it easy to delete empty namespaces on the agent side?

dhirajsb:
We should add another property to the namespace with a count of connectors deployed e.g. `connectors_count`. This will be different from the status `deployed_connectors` which represents connector CRs. 

So, if the namespace phase is `deleting` and this new property `connectors_count` is 0, then fleet shard can go ahead and delete the namespace. When that namespace is not in the list of namespaces in the cluster status update in the future, fleetmanager will internally change the state to deleted, and delete the namespace row in the DB. 

wdyt?

lburgazzoli:
> We should add another property to the namespace with a count of connectors deployed e.g. `connectors_count`. This will be different from the status `deployed_connectors` which represents connector CRs.
> 

Done in https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/pull/969
Note that this is an interim solution and should be replaced by a dedicated resource.

lburgazzoli:
@dhirajsb I think we don't need the annotation but since the status is managed by the manager, then the manager must also set the connectors count based on what it is in the database and do not propagate what the fleetshard provides (same as for the status).

dhirajsb:
@lburgazzoli I think it's better to keep the annotation, as we're planning on making fleetshard owner of the status in the future. So we'll end up redoing the status field hack if we go that way now. The new annotation is readonly, and user is not allowed to set or change it. 

Let's keep the status `connectors_deployed` count to be the number of connector CRs in the fleetshard. 
It also makes it easy to compare the number from annotations with the number in the status to see whether the agent has deployed all the assigned connectors. 

wdyt?

dhirajsb:
If you are worried about exposing the new annotation to end users, fleetmanager can be changed to only send it to agents.

lburgazzoli:
> @lburgazzoli I think it's better to keep the annotation, as we're planning on making fleetshard owner of the status in the future. So we'll end up redoing the status field hack if we go that way now. 
> 

In the future we need to have a dedicated resource, like for connectors we have  connector deployments so we can set a desired state and any other information the shard would require to operator without having to leak them to the public APIs so the shard won't own the namespace status as it is for connectors.

> Let's keep the status `connectors_deployed` count to be the number of connector CRs in the fleetshard. It also makes it easy to compare the number from annotations with the number in the status to see whether the agent has deployed all the assigned connectors.
> 

In such case, would't ti be better to have a dedicates field with the number of connectors in the namespace instead of using an annotation ? that would allow to keep the connectors_deployed but also having the count of the number of connectors in the namespace.

dhirajsb:
> In the future we need to have a dedicated resource, like for connectors we have connector deployments

Even with a dedicated resource in between like a `namespace_deployment`, the namespace status will still be copied from that and owned by fleetshard just like we do for connectors and connector deployments now. 

> would't ti be better to have a dedicates field with the number of connectors in the namespace instead of using an annotation

I thought we didn't want to make it part of the spec because it's a computed dynamic value and not really part of the spec used to create the namespace. Hence we ended up deciding on using an annotation.

lburgazzoli:
> > In the future we need to have a dedicated resource, like for connectors we have connector deployments
> 
> Even with a dedicated resource in between like a `namespace_deployment`, the namespace status will still be copied from that and owned by fleetshard just like we do for connectors and connector deployments now.
> 

We don't need to blindly copy it as we don't blindly copy it as today for connectors.

> > would't ti be better to have a dedicates field with the number of connectors in the namespace instead of using an annotation
> 
> I thought we didn't want to make it part of the spec because it's a computed dynamic value and not really part of the spec used to create the namespace. Hence we ended up deciding on using an annotation.

We should put it in the **status**

rinaldodev (approved):

dhirajsb:
So, does this PR assume the fleet manager is going to send connector count in `status.connectors_deployed`??

lburgazzoli:
> So, does this PR assume the fleet manager is going to send connector count in `status.connectors_deployed`??

Yes, this is the current behavior (https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/openapi/connector_mgmt.yaml#L1437-L1451)

dhirajsb:
Ok, that was originally proposed to be data plane side CR count and hence was being persisted in the DB. But I guess that's been changed now. 
So I'll modify the fleetmanager, and ignore any writes to the DB. I won't be making any changes to the underlying DB schema this close to preview release. So that column will be ignored, because having to maintain and sync that field otherwise will add unnecessary complexity to fleetmanager.

PR #352, opened 2022-05-04
MGDCTRS-901 Configure database history for eligible connectors

Naros:
https://issues.redhat.com/browse/MGDCTRS-901

Automatically configures the database history topic as "{connector-name}-database-history" for Debezium connectors that require a database history topic configuration.  This change should allow MySQL and SQL Server connectors to work correctly since they require a database history configuration.  Future connectors such as Db2 and Oracle will also require this.

### DIFF
--- a/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
+++ b/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
@@ -2,7 +2,9 @@ package org.bf2.cos.fleetshard.operator.debezium;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
```diff
@@ -206,6 +209,28 @@ protected List<HasMetadata> doReify(
...
+            databaseHistoryConfigs.put("database.history.kafka.bootstrap.servers",
+                connector.getSpec().getDeployment().getKafka().getUrl());
+            databaseHistoryConfigs.put("database.history.kafka.topic", connector.getMetadata().getName() + "-database-history");
+            databaseHistoryConfigs.put("database.history.producer.security.protocol", "SASL_SSL");
+            databaseHistoryConfigs.put("database.history.consumer.security.protocol", "SASL_SSL");
+            databaseHistoryConfigs.put("database.history.producer.sasl.mechanism", "PLAIN");
+            databaseHistoryConfigs.put("database.history.consumer.sasl.mechanism", "PLAIN");
+            databaseHistoryConfigs.put("database.history.producer.sasl.jaas.config",
+                "org.apache.kafka.common.security.plain.PlainLoginModule required username=\""
+                    + serviceAccountSpec.getClientId() + "\" password=\"" + password + "\";");
```
lburgazzoli, line 226: In general, password should not be part of a CR but should be referenced from a secret (i.e. https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/0475269fa7989c4d6b07ad2ecdea9272653aef96/cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandSupport.java#L100-L107)

lburgazzoli (changes requested):
Secrets should not be part of the CR and we should have a test for this

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
Naros: I'll take a look.  I tried to use the secrets configuration reader to pass a reference to the client password from another existing secret but that didn't seem to work, unfortunately. I'll verify if we can use the file reader here.

Naros: Found a workable solution, checking to verify that its read properly by Kafka Connect.

lburgazzoli:
@Naros looks good, can we add a small test ?

lburgazzoli (approved):

PR #356, opened 2022-05-09
Suport namespace quota and status reporting

lburgazzoli:
- MGDCTRS-828: set resource quota assigned to a kubernetes namspace
- MGDCTRS-827: update the status of namespaces when there is a failure creating the resource on kubernetes

### DIFF
--- a/cos-fleetshard-operator-camel-it/pom.xml
+++ b/cos-fleetshard-operator-camel-it/pom.xml
@@ -50,6 +50,8 @@
                         <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
                         <maven.home>${maven.home}</maven.home>
                     </systemPropertyVariables>
+                    <forkCount>1</forkCount>
+                    <reuseForks>false</reuseForks>
                 </configuration>
             </plugin>
             <plugin>
--- a/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorNamespaceProvisioner.java
+++ b/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorNamespaceProvisioner.java
@@ -2,6 +2,7 @@ package org.bf2.cos.fleetshard.sync.resources;
 
 import java.util.Collection;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 
 import javax.enterprise.context.ApplicationScoped;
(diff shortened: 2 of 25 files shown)


rinaldodev (approved):

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorNamespaceProvisioner.java
```diff
@@ -160,30 +236,49 @@ public void provision(ConnectorNamespace namespace) {
...
         Resources.setAnnotations(
             ns,
-            Resources.ANNOTATION_NAMESPACE_EXPIRATION, namespace.getExpiration());
+            Resources.ANNOTATION_NAMESPACE_EXPIRATION, connectorNamespace.getExpiration(),
+            Resources.ANNOTATION_NAMESPACE_QUOTA, Boolean.toString(hasQuota(connectorNamespace)));
 
         fleetShard.createNamespace(ns);
 
-        copyAddonPullSecret(ns);
+        if (hasQuota(connectorNamespace)) {
```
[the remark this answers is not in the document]
lburgazzoli: +1

cos-fleetshard-operator-camel-it/pom.xml
```diff
@@ -50,6 +50,8 @@
                         <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
                         <maven.home>${maven.home}</maven.home>
                     </systemPropertyVariables>
+                    <forkCount>1</forkCount>
```
rk3rn3r, line 53: I don't think this has any effect as it is the default, does it? I think it even yields a notification that this has no effect and considered a more or less invalid config?

lburgazzoli: I have not noticed any warning but I may have missed it. 

I usually explicitly set values to avoid inherited values but I could remove it.

rk3rn3r: I just remember some issue with that value recently with the DBZ UI. Maybe @Naros  remembers when we hit that?

lburgazzoli: maven does not seem to complain

```
[INFO] --- maven-surefire-plugin:3.0.0-M6:test (default-test) @ cos-fleetshard-sync-it ---
[INFO] Using auto detected provider org.apache.maven.surefire.junitplatform.JUnitPlatformProvider
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
```

rinaldodev (approved):
lgtm

cos-fleetshard-operator-camel-it/pom.xml
lburgazzoli: @rk3rn3r @Naros is shit a blocking issue ? or we can address it in a follow up PR if needed ?

The reason it has been added here is to be sure tests are executed in a new jvm since otherwise this could lead to OOM exception on GitHub Actions.

rinaldodev (approved):

PR #364, opened 2022-05-23
Set limit ranges assigned to a kubernetes namwspace

lburgazzoli:
With this PR a new `LimitRange` resource is created in any namespace where resource quota are configured assuming that at this stage, only eval namespace have such quota set.
 
The reason of such resource to be created is that,  when allocating compute resources, each container may specify a request and a limit value for either CPU or memory and the quota can be configured to quota either value per namespace but if the quota has a value specified for requests.cpu or requests.memory, then it requires that every container makes an explicit request for those resources. If the quota has a value specified for limits.cpu or limits.memory, then it requires that every incoming container specifies an explicit limit for those resources.

Since not our connector do explicit specify limits, a `LimitRange` can be used to automatically set them:

```yaml
apiVersion: v1
kind: LimitRange
metadata:
  name: redhat-openshift-connectors-ca5j6vufrusf747mmmmg-ranges
spec:
  limits:
  - default:
      cpu: 500m
      memory: 500m
    defaultRequest:
      cpu: 200m
      memory: 128m
    type: Container
```

With this set-up each new container that is created on the eval namespace, will receive a default limit computed taking into account the limits sets on the namespace and the number of connectors that are allowed to be created to the namespace.

### DIFF
--- a/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorNamespaceProvisioner.java
+++ b/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorNamespaceProvisioner.java
@@ -23,6 +23,9 @@ import org.bf2.cos.fleetshard.sync.client.FleetShardClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.fabric8.kubernetes.api.model.LimitRange;
+import io.fabric8.kubernetes.api.model.LimitRangeItem;
+import io.fabric8.kubernetes.api.model.LimitRangeSpec;
 import io.fabric8.kubernetes.api.model.Namespace;
 import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.Quantity;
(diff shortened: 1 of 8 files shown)


cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorNamespaceProvisioner.java
```diff
@@ -203,6 +210,76 @@ private void createResourceQuota(String uow, ConnectorNamespace connectorNamespa
...
+        if (connectorNamespace.getQuota() == null) {
+            return;
+        }
+
+        LimitRangeItem limit = new LimitRangeItem();
+        limit.setType(LIMITS_TYPE_CONTAINER);
+        limit.setDefault(new TreeMap<>());
+        limit.setDefaultRequest(new TreeMap<>());
+
+        if (config.quota().defaultLimits() != null) {
```
lburgazzoli, line 223: The code above computes the limits based on the number of connectors and the limits set on the namespace. 

However there may be the case where a connector declares requests that could be above the limits this method would compute, however, since `LimitRange` kicks in only in case of missing requests/limits, a connector could set higher limits.

@rinaldodev what do you think?

valdar (approved):

PR #369, opened 2022-05-24
The Fleetshard Sync must retrieve the SSO base URL from api.openshift.com

lburgazzoli:
MAS APIs exposes an endpoint (/api/kafkas_mgmt/v1/sso_providers) to get information
about the SSO provider for the service, as example:

```json
{
  "value": {
    "base_url": "https://identity.api.redhat.com",
    "token_url": "https://identity.api.openshift.com/auth/realms/rhoas/protocol/openid-connect/token",
    "valid_issuer": "https://identity.api.openshift.com/auth/realms/rhoas"
  }
}
```

With this commit, the cos-fleetshard-sync does not require anymore to explicitly set the
address of the SSO as it leverages the sso_providers endpoint to retrieve the required
information to configure the OidcClient.


heiko-braun:
Is the idea still to perform the check periodically? I guess we will not migrate SSO every other week, so eventually just doing it at startup would suffice?

lburgazzoli:
> Is the idea still to perform the check periodically? I guess we will not migrate SSO every other week, so eventually just doing it at startup would suffice?

That is the behavior described in the issue, should we do it only once ?

heiko-braun:
Well, peridoc checks help duirng the SSO migration, but the interval needs to be short to help with that. If it's too long, the data planes might be defunct until they retrieve the next SSO URL and token. After that point, once the migration is complete, it basically becomes a periodic waste of CPU cycles. Which might not be a big problem.  Food for thought – at the end it's your decision.

lburgazzoli:
> Well, peridoc checks help during the SSO migration, but the interval needs to be short to help with that. If it's too long, the data planes might be defunct until they retrieve the next SSO URL and token. After that point, once the migration is complete, it basically becomes a periodic waste of CPU cycles. Which might not be a big problem. Food for thought – at the end it's your decision.

Got it. I'll try to implement a better solution then

lburgazzoli:
@heiko-braun refactored to lazily retrieve the sso url and to re-fresh it every time a communication error occours

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/client/AuthRequestFilter.java
```diff
@@ -0,0 +1,77 @@
...
+                tokenSupplier.reset();
+            });
+    }
+
+    @Override
+    public void filter(ClientRequestContext requestContext, ClientResponseContext responseContext) throws IOException {
+        final int code = responseContext.getStatus();
+        final boolean success = code / 100 == 2;
+
+        if (!success && code != 410) {
```
rinaldodev, line 73: We seem to be recreating the client on everything different from 2xx or 410. Is this intended?

lburgazzoli: Yes as for some auth issues, the error is just 404 or a 500x but we don't know exactly if it is an issue with the token or what not. 

Do you think there is anything better ?

lburgazzoli: @rinaldodev fixed

abkieling:
@lburgazzoli Retrieving the SSO URL when the communication fails won't work during the SSO migration. It won't fail because both old and new SSO URLs will work during the migration window (4 hours). We need an interval configuration that can be set to a smaller value during the migration day (e.g., 30 minutes) and a larger value (or zero meaning disabled) after the migration.

lburgazzoli:
> @lburgazzoli Retrieving the SSO URL when the communication fails won't work during the SSO migration. It won't fail because both old and new SSO URLs will work during the migration window (4 hours). We need an interval configuration that can be set to a smaller value during the migration day (e.g., 30 minutes) and a larger value (or zero meaning disabled) after the migration.

I can add such logic as in fact, it was present in one of the earlier PR, however we would need to deploy a new bundle to change the configuration and a new one to reset it back so, as we need to roll-out a new version anyway, I wonder if we leverage the sync restart as consequence of the upgrade as a trigger for the re-configuration.

@heiko-braun @rinaldodev @abkieling WDYT ?

rinaldodev:
@lburgazzoli @heiko-braun @abkieling I'd say to add the frequent update for now, as an addition to the error detection, and later change that to a not-so-frequent one (a few hours maybe).

lburgazzoli:
@abkieling @rinaldodev url refresh has been implemented (default 1h), mind having a look ?

rinaldodev (approved):

abkieling:
@rinaldodev @lucamolteni I have been informed today the API endpoint that returns the SSO URL is temporary. After the migration, we will need to change the cos-fleetshard-sync again and go back to an add-on parameter or just hard-code the URL.

rinaldodev:
@abkieling talked with @lburgazzoli and we will merge and remove this afterwards.

PR #391, opened 2022-06-11
MGDCTRS-1057: Temporary disable exchange pooling as it has side effects on some connectors

lburgazzoli:


cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandConfiguration.java
```diff
@@ -63,7 +63,13 @@ interface Health {
...
 
     interface ExchangePooling {
-        @WithDefault("pooled")
+        // The pooled exchange factory strategy has side effects on some components:
+        // - https://issues.redhat.com/browse/MGDCTRS-1058
+        // - https://issues.redhat.com/browse/MGDCTRS-1057
+        // - https://issues.apache.org/jira/browse/CAMEL-18187
+        //
+        // Default changed to "prototype" till we get a fix in Apache Camel
+        @WithDefault("prototype")
```
lburgazzoli, line 72: changed the default instead of using a config option because given the issue, `prototype` seems to be a sensible default for now

rinaldodev: 👍

rinaldodev (approved):

PR #410, opened 2022-07-12
MGDCTRS-1139 Handle DebeziumExceptions that lead to Kafka Connect task errors but not fails the connector

rk3rn3r:
closes https://issues.redhat.com/browse/MGDCTRS-1139

### DIFF
--- a/cos-fleetshard-operator-debezium-it/src/test/java/org/bf2/cos/fleetshard/operator/it/debezium/glues/KafkaConnectorSteps.java
+++ b/cos-fleetshard-operator-debezium-it/src/test/java/org/bf2/cos/fleetshard/operator/it/debezium/glues/KafkaConnectorSteps.java
@@ -29,12 +29,12 @@ import static org.bf2.cos.fleetshard.it.cucumber.assertions.CucumberAssertions.a
 public class KafkaConnectorSteps extends StepsSupport {
     @Then("the kctr exists")
     public void exists() {
-        awaiter.until(() -> kctr() != null);
+        awaiter.until(() -> kafkaConncetor() != null);
     }
 
     @Then("the kctr does not exists")
     public void does_not_exists() {
-        awaiter.until(() -> kctr() == null);
+        awaiter.until(() -> kafkaConncetor() == null);
     }
 
     @When("the kctr path {string} is set to json:")
@@ -52,7 +52,7 @@ public class KafkaConnectorSteps extends StepsSupport {
 
     @And("the kctr has an entry at path {string} with value {string}")
     public void kctr_has_a_path_matching_value(String path, String value) {
-        KafkaConnector res = kctr();
+        KafkaConnector res = kafkaConncetor();
 
         assertThat(res)
             .isNotNull();
@@ -64,7 +64,7 @@ public class KafkaConnectorSteps extends StepsSupport {
 
     @And("the kctr has an entry at path {string} with value {int}")
     public void kctr_has_a_path_matching_value(String path, int value) {
-        KafkaConnector res = kctr();
+        KafkaConnector res = kafkaConncetor();
 
         assertThat(res)
             .isNotNull();
@@ -76,7 +76,7 @@ public class KafkaConnectorSteps extends StepsSupport {
 
     @And("the kctr has an entry at path {string} with value {bool}")
     public void kctr_has_a_path_matching_value(String path, Boolean value) {
-        KafkaConnector res = kctr();
+        KafkaConnector res = kafkaConncetor();
 
         assertThat(res)
             .isNotNull();
@@ -88,7 +88,7 @@ public class KafkaConnectorSteps extends StepsSupport {
 
     @And("the kctr has an object at path {string} containing:")
     public void kctr_has_a_path_matching_object(String path, String content) {
-        KafkaConnector res = kctr();
+        KafkaConnector res = kafkaConncetor();
 
         assertThat(res)
             .isNotNull();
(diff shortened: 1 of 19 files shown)


rk3rn3r:
@lburgazzoli @Naros @rinaldodev this PR is finally ready for review!

cos-fleetshard-operator-debezium-it/src/test/java/org/bf2/cos/fleetshard/operator/it/debezium/glues/KafkaConnectorSteps.java
```diff
@@ -176,6 +176,33 @@ public void kctr_status_is(String phase) {
                 return resource;
             });
     }
+    @When("the kctr status is RUNNING with a failed task")
+    public void kctr_with_failed_task() {
```
lburgazzoli, line 180: Not an issue but wouldn't be better to make this step configurable ? so you can write something like:

```gherkin
 When the kctr is "RUNNING" with tasks status:
   | id | state  | trace                       |
   | 0  | FAILED | The connector is trying ... |
```

rk3rn3r: Not really. The message is superfluous and doesn't have to be customized. I might revisit this when we look into multi-task connectors.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandSupport.java
```diff
@@ -176,99 +182,160 @@ private static io.fabric8.kubernetes.api.model.Condition cloneCondition(
...
-        readyCondition.setReason("Transitioning");
-
-        boolean kafkaConnectorFailed = true;
+    private static void computeConnectorStatus(
+        KafkaConnector kafkaConnector,
+        ConnectorStatus connectorStatus) {
 
         if (null != kafkaConnector) {
+            Condition connectorReadyCondition = null;
+            Condition connectorNotReadyCondition = null;
```
lburgazzoli, line 191: It is needed to have both ? The Ready and NotReady conditions can be probably collapsed into a single Ready condition where the status is true or false, which could simplify some of the flow control later in this method

cos-fleetshard-operator-debezium-it/src/test/java/org/bf2/cos/fleetshard/operator/it/debezium/glues/KafkaConnectorSteps.java
lburgazzoli: Ok, since there were a very specific error message I thought it was meaningful and that such error was the one to catch. Maybe add a comment to say that error is not relevant

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandSupport.java
rk3rn3r: That's true, but they need to be and are combined later. We never know in which order they are coming in the list of conditions and we need to capture both first while iterating over the list of conditions. Later they are aggregated into one readyCondition. It even gets combined with the KafkaConnect conditions later on.

lburgazzoli: AFAIK, I don't think you can have both Ready and NotReady condition at the same time in fact, there's no Ready/False and NotReady/True (at least last time I checked)

rk3rn3r: @lburgazzoli can you link to where this can be checked?

lburgazzoli: I was looking at this code: https://github.com/strimzi/strimzi-kafka-operator/blob/56342fd3d3f0c86e54e0d44a24474856a2af43e2/operator-common/src/main/java/io/strimzi/operator/common/operator/resource/StatusUtils.java#L92-L104 so my understanding is that the "ready" condition can in fact be either "Ready" or "NotReady".

It is not a common pattern in Kubernetes but it make sense that they are mutually exclusive i.e. can you have a condition for which a resource is both ready and not ready at the same time ?

Maybe it is worth checking with the strimzi dev

rk3rn3r: I expected `{"Ready": "False", "NotReady": "True"}` in some situations. There might even be timing issues where `{"Ready": "True", "NotReady": "True"}` co-exist, even future bugs? When I tried to figure this out I found it very complex to figure out when which condition is set and removed.
The current implementation is resilient to any situation and I would just leave the code as it is and would prefer to not spend even more time on this. I wasted enough time to make the integration tests work.

lburgazzoli: According to the code, the conditions you described should not be possible so I would recommend to get in touch with the strimzi team to have some understanding of what is expected but that's your call.

lburgazzoli (approved):

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandSupport.java
rk3rn3r: I still lean towards this more robust and resilient implementation here. If we don't see such situations in production for a while I am okay with revisiting this part of the code.

PR #413, opened 2022-07-19
MGDCTRS-1177: Create Subscription and copy Secrets and ConfigMaps

rinaldodev:


lburgazzoli (changes requested):
We should disable observability and subscription resource copy in dev/local overlay as it would fail as there's no bundle in dev/local and also it may fail in case the observability resource and related CRDs are not installed.

rinaldodev:
> We should disable observability and subscription resource copy in dev/local overlay as it would fail as there's no bundle in dev/local and also it may fail in case the observability resource and related CRDs are not installed.

It was already disabled in the [local overlay](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/main/etc/kubernetes/manifests/overlays/local/cos-fleetshard-sync/app-config/application.properties#L19). Added the property in dev as well.

etc/kubernetes/manifests/overlays/local/observability/crds/customresourcedefinition-subscriptions.operators.coreos.com.yaml
```diff
@@ -0,0 +1,2636 @@
+﻿kind: CustomResourceDefinition
```
lburgazzoli, line 1: is is correct that this CR is in the local overlay ?

etc/kubernetes/manifests/overlays/local/observability/crds/customresourcedefinition-subscriptions.operators.coreos.com.yaml
```diff
@@ -0,0 +1,2636 @@
...
+  name: subscriptions.operators.coreos.com
+  uid: 2442059f-ce23-4d68-8f6d-1c9a6d1c0b6b
+  resourceVersion: '1947'
+  generation: 1
+  creationTimestamp: '2022-05-30T13:00:00Z'
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.6.2
+    include.release.openshift.io/ibm-cloud-managed: 'true'
+    include.release.openshift.io/self-managed-high-availability: 'true'
+  ownerReferences:
```
lburgazzoli, line 13: looks like this is an export of a CR, is it correct ?

etc/scripts/deploy_observability_crds.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash -e
 
-kubectl apply -f etc/kubernetes/manifests/overlays/local/observability/crds/observabilities.observability.redhat.com-v1.yml 
+kubectl apply -f etc/kubernetes/manifests/overlays/local/observability/crds/observabilities.observability.redhat.com-v1.yml
```
lburgazzoli, line 3: is it correct that such resources are from the local overlay ?

lburgazzoli: ah ok, maybe we can have an it overlay and we can just do an kubectl apply-k of the overlay to get the it set-up, what do  you think ?

etc/kubernetes/manifests/overlays/local/observability/crds/customresourcedefinition-subscriptions.operators.coreos.com.yaml
rinaldodev: Yes it was exported. Cleaned.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/client/FleetShardObservabilityClient.java
```diff
@@ -0,0 +1,216 @@
...
+            LOGGER.error("Observability CustomResourceDefinition is not present. Observability will not be enabled.");
+            return;
+        }
+
+        if (observabilityNamespace.isBlank()) {
+            LOGGER.error("Observability namespace is not set!");
+            return;
+        }
+
+        if (config.observability().subscription().enabled()) {
```
lgarciaaco, line 65: What's the use case where a user sets `enabled` to `true` and `subscription.enabled` to false?

lgarciaaco (approved):
With some comments but all in all fine

etc/scripts/deploy_observability_crds.sh
rinaldodev: @lburgazzoli doing in this one

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/client/FleetShardObservabilityClient.java
rinaldodev: If/when we are able to tell the observability-operator to install the prometheus operator in a different namespace. Then we can (not mandatory) just disable the Subscription and install as a dependency.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/client/FleetShardObservabilityClient.java
```diff
@@ -0,0 +1,216 @@
...
+            createSubscriptionResource();
+        }
+        copyResourcesToObservabilityNamespace();
+        createObservabilityResource();
+    }
+
+    private void createSubscriptionResource() {
+        LOGGER.info("Creating Subscription resource");
```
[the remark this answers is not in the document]
rinaldodev: Currently in dev/local observability is disabled, so it won't have an impact right now. I would also consider creating the namespace in the kustomize overlays. That said I'd leave that for later. WDYT?

lburgazzoli (approved):

PR #446, opened 2022-08-23
feat: improve generated kubernetes resources sanitization scripts

lburgazzoli:
This PR is aimed at reducing the noise caused by changes in the generated kubernetes resources (CRDs, manifests, etc)


rinaldodev (approved):
Very helpful, thanks. LGTM.
Would be nice to have it documented somewhere, maybe even in the pom.xml.

lburgazzoli:
> Very helpful, thanks. LGTM. Would be nice to have it documented somewhere, maybe even in the pom.xml.

Will add some doc in a follow up PR

PR #452, opened 2022-08-30
MGDCTRS-1431: send k8s events on addon removal

rinaldodev:


cos-fleetshard-support/src/main/java/org/bf2/cos/fleetshard/support/client/EventClient.java
```diff
@@ -0,0 +1,120 @@
...
+        EventType(String text) {
+            this.text = text;
+        }
+
+        public String getText() {
+            return text;
+        }
+
+        public static Optional<EventType> fromText(String text) {
+            return Arrays.stream(values()).filter(e -> e.getText().equals(text)).findFirst();
```
lburgazzoli, line 43: This is not going to be an issue given that this method is probably not invoked many time, however this produces a number of temporary objects, i.e:
- invoking `values()` generates a new array every time it is invoked and such array it is often cached as a static field
- it allocates a stream and the related internal objects

Most of such temporary allocation can be avoided by something like:

```java
public enum EventType {
    static final EventType[] VALS = values();
    
    public static Optional<EventType> fromText(String text) {
        for (int i = 0; i < val.lenght; i++} {
            if (VALS[i].getText().equals(text)) {
                return Optional.of(VALS[i]);
            }
            
            return Optional.empty();
        }
    }
}
```

lburgazzoli (approved):
Just a minor note but LGTM

lburgazzoli (approved):

PR #461, opened 2022-09-15
Added Changes to expose connector state metrics

shivamG640:
Hi @lburgazzoli, @rinaldodev
Please have a look at this PR. This is regarding exposing the connector state metric from the data plane!

Regards, Shivam

### DIFF
--- a/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
+++ b/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
@@ -3,21 +3,37 @@ package org.bf2.cos.fleetshard.sync.resources;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import org.bf2.cos.fleet.manager.model.ConnectorDeploymentStatus;
 import org.bf2.cos.fleetshard.api.ManagedConnector;
+import org.bf2.cos.fleetshard.sync.FleetShardSyncConfig;
 import org.bf2.cos.fleetshard.sync.client.FleetManagerClient;
 import org.bf2.cos.fleetshard.sync.client.FleetManagerClientException;
 import org.bf2.cos.fleetshard.sync.client.FleetShardClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
 @ApplicationScoped
 public class ConnectorStatusUpdater {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConnectorStatusUpdater.class);
+
+    public static final String CONNECTOR_STATE = "connector.state";
+    private List<Tag> tags;
+
 
     @Inject
     FleetManagerClient fleetManagerClient;
     @Inject
     FleetShardClient connectorClient;
+    @Inject
+    MeterRegistry registry;
+    @Inject
+    FleetShardSyncConfig config;
 
     public void update(ManagedConnector connector) {
         LOGGER.debug("Update connector status (name: {}, phase: {})",
@@ -25,9 +41,31 @@ public class ConnectorStatusUpdater {
             connector.getStatus().getPhase());
 
         try {
-            fleetManagerClient.updateConnectorStatus(
-                connector,
-                ConnectorStatusExtractor.extract(connector));
+            ConnectorDeploymentStatus connectorDeploymentStatus = ConnectorStatusExtractor.extract(connector);
+
+            fleetManagerClient.updateConnectorStatus(connector, connectorDeploymentStatus);
+
+            LOGGER.debug("Updating change status metrics (Connector_id : {}, state: {})",
+                connector.getSpec().getConnectorId(),
+                ConnectorStatusExtractor.extract(connector).getPhase());
+
+            switch (connectorDeploymentStatus.getPhase()) {
+                case READY:
+                    measure(connector, 1);
+                    break;
+                case FAILED:
+                    measure(connector, 2);
+                    break;
+                case DELETED:
+                    measure(connector, 3);
+                    break;
+                case STOPPED:
+                    measure(connector, 4);
+                    break;
+                default:
+                    measure(connector, 5);
+                    break;
+            }
 
         } catch (FleetManagerClientException e) {
             if (e.getStatusCode() == 410) {
@@ -43,4 +81,23 @@ public class ConnectorStatusUpdater {
         }
     }
 
+    /* Expose a metric "cos_fleetshard_sync_connector_state" which reveals the current connector state.
+    Metric value of 1 implies that the connector is in Ready state. Similarly, 2 -> Failed, 3 -> Deleted,
+    4 -> Stopped, 5 -> In Process */
+    private void measure(ManagedConnector connector, Integer connectorState) {
+
+        tags = List.of(
+                Tag.of("cos.connector.id", connector.getSpec().getConnectorId()),
+                Tag.of("cos.deployment.id", connector.getSpec().getDeploymentId()));
+
+        Gauge gauge = registry.find(config.metrics().baseName() + "." + CONNECTOR_STATE).tags(tags).gauge();
+        if (gauge != null) {
+            registry.remove(gauge);
+        }
+
+        Gauge.builder(config.metrics().baseName() + "." + CONNECTOR_STATE, new AtomicInteger(connectorState), AtomicInteger::get)
+            .tags(tags)
+            .register(registry);
+
+    }
 }


cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -43,4 +81,23 @@ public void update(ManagedConnector connector) {
         }
     }
 
+    /* Expose a metric "cos_fleetshard_sync_connector_state" which reveals the current connector state.
+    Metric value of 1 implies that the connector is in Ready state. Similarly, 2 -> Failed, 3 -> Deleted,
+    4 -> Stopped, 5 -> In Process */
+    private void measure(ManagedConnector connector, Integer connectorState) {
+
+        tags = List.of(
```
lburgazzoli, line 89: you must not use an instance variable here as if this method get invoked my multiple threads, then you may get unpredictable results

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -43,4 +81,23 @@ public void update(ManagedConnector connector) {
...
+                Tag.of("cos.connector.id", connector.getSpec().getConnectorId()),
+                Tag.of("cos.deployment.id", connector.getSpec().getDeploymentId()));
+
+        Gauge gauge = registry.find(config.metrics().baseName() + "." + CONNECTOR_STATE).tags(tags).gauge();
+        if (gauge != null) {
+            registry.remove(gauge);
+        }
+
+        Gauge.builder(config.metrics().baseName() + "." + CONNECTOR_STATE, new AtomicInteger(connectorState), AtomicInteger::get)
```
lburgazzoli, line 98: This may be simplified by using `builder(String name, Supplier<Number> f)` where the supplied can be cached

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -43,4 +81,23 @@ public void update(ManagedConnector connector) {
...
+    4 -> Stopped, 5 -> In Process */
+    private void measure(ManagedConnector connector, Integer connectorState) {
+
+        tags = List.of(
+                Tag.of("cos.connector.id", connector.getSpec().getConnectorId()),
+                Tag.of("cos.deployment.id", connector.getSpec().getDeploymentId()));
+
+        Gauge gauge = registry.find(config.metrics().baseName() + "." + CONNECTOR_STATE).tags(tags).gauge();
+        if (gauge != null) {
+            registry.remove(gauge);
```
lburgazzoli, line 95: can you explain the reason to remove the Gauge ?

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -3,31 +3,69 @@
...
+
+            fleetManagerClient.updateConnectorStatus(connector, connectorDeploymentStatus);
+
+            LOGGER.debug("Updating change status metrics (Connector_id : {}, state: {})",
+                connector.getSpec().getConnectorId(),
+                ConnectorStatusExtractor.extract(connector).getPhase());
+
+            switch (connectorDeploymentStatus.getPhase()) {
+                case READY:
+                    measure(connector, 1);
```
lburgazzoli, line 54: don't use magic values like 1,2,3,4,5, constants would be better or alternatively the PhaseType enum can be enhanced to include additional values.

lburgazzoli (changes requested):

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
shivamG640: This is because it's difficult to get the reference of the already registered gauge to update its value when the connector state change. Found out that this way of removing the registered guage and registering a new one here: 
https://github.com/micrometer-metrics/micrometer/issues/2024

lburgazzoli: butbif you find it, can't you update the value ? son you would create it only once and any further call would just update it.

I'm mainly concern that this creates a number of temporary object that would - maybe marginal - impact the garba collection

shivamG640:
Hi @lburgazzoli ,

Are the changes good enough to get merged?
I am not sure how to change the value of the gauge so I am deleting and recreating it. As we discussed, I will raise an issue on this to fix this further.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
lburgazzoli: I found this issue https://github.com/micrometer-metrics/micrometer/issues/480 that may be something we want to do in order to avoid registering and deleting a gauge every time. 

Maybe you can register a MutableGauge and use it afterward

rinaldodev: @lburgazzoli I guess always building a new one is what we already do with our Timers and Counters, since we don't hold pointers to the instances.
https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/d74ed9710b8270df67c892481d12d2e3c8a8c94c/cos-fleetshard-support/src/main/java/org/bf2/cos/fleetshard/support/metrics/MetricsRecorder.java#L61-L71

We could use MeterRegistry.gauge, or .timer, or .counter, since those internally use `registerMeterIfNecessary`, saving some performance. But we should probably do it to all of them.

lburgazzoli: yep

rinaldodev: The key take-away being: I vote we move on with this current state and open an issue for the enhancement.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
rinaldodev: Agreed. I'd rather we use `connector.getStatus().getPhase()` for reporting, and augmenting the PhaseType to hold the enum values instead of constants that don't belong to that class.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
lburgazzoli: fine with me, can you open an issue ?

shivamG640: @rinaldodev, AFAIK, we do build new Timers and Counters always but it internally does a check and registers new metrics only if the metrics name+tags is a unique combination. If there is an existing metric with a name+tags combination, It overrides its value.
But gauge works differently, it only registers metrics with a unique set of name+tags combination and does not updates its value if there is an existing one.

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
lburgazzoli: @rinaldodev is there any change related to this you want before merging ?

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
rinaldodev: @shivamG640 ah, yes, you are correct, sorry I haven't read the issue you pointed before and misunderstood the problem. 

@lburgazzoli we don't need to create the issue, and the MutableGauge was just an idea from micrometer, it's not implemented yet.

I have investigated a little bit more, and [the docs clear state](https://micrometer.io/docs/concepts#_gauges:~:text=It%20is%20your%20responsibility%20to%20hold%20a%20strong%20reference%20to%20the%20state%20object%20that%20you%20are%20measuring%20with%20a%20Gauge.) that when you use a Gauge, you are responsible for maintaining a reference to the object you are measuring (the atomic int in our case). There is also the `strongReference` option, which is implemented but not yet documented, and it instructs the Gauge to hold the reference being measured, but then again it doesn't provide a way to _retrieve_ the reference and update it's value, so it's doesn't help in this specific case.

That said, I can see 3 options (there could be more):

1. keep the Gauge creation where it is in this PR, updating the state of the connector the same time we call the fleet-manager, and use the current workaround of deleting and recreating the gauge.

1. keep the Gauge creation where it is in this PR, updating the state of the connector the same time we call the fleet-manager, but keep references to the `AtomicInt`s and update their value when needed. This is probably simpler and performatic, just not so pretty.

1. move the Gauge creation to the same time we create a new MannagedConnector, and in the measuring function we use the `connectorsInformer` to query the connector and extract its current state. That would provide the most up to date information, use the in-memory cache we already have in the informer, but maybe some extra cpu usage over option 2.

But I'm fine with doing all of this improvements later.

lburgazzoli: @shivamG640 can you open an issue for the points mentioned by @rinaldodev ? we can discuss later and move on with thine one.

lburgazzoli (approved):

rinaldodev (approved):

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
shivamG640: @rinaldodev, Sorry I didn't see this comment the last time. I agree with you that we should introduce int values with PhaseType instead of constants.
My suggestion will be to also re-evaluate both ConnectorState and PhaseType enums and keep only one such type that confirms the current status/state of the connector (if they don't provide a different kind of information about the connector).

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
shivamG640: Thanks, @rinaldodev, and @lburgazzoli. Created https://issues.redhat.com/browse/MGDCTRS-1520 to discuss this further!

PR #480, opened 2022-10-06
MGDCTRS-1132: Connectors stuck in deleting state

lburgazzoli:
This PR add some workarounds - `Syste.exit()¡ in case of unrecoverable errors - to improve the resilience of the data plane in case the communication with the Kubernetes API server fails silently causing the operators and the sync to become stale. 

In order to be notified of unrecoverable errors, the following dependencies have been upgraded as the fix has been introduced recently [2]

- JOSDK 3.2.3
- Fabric8 client 5.12.4 
- Quarkus 2.13.1

Unfortunately there is a side effect on the automatic generation of some kubernetes resources  manifest generation so the selector for `ServiceMonitor` are not automatically added hence I had manually configure them.

Notes:
- is is not easy to reproduce this particular issue so there's no test at the moment, I will follow up with some tests as soon as I have a good way to test it

[1] https://issues.redhat.com/browse/MGDCTRS-1132
[2] https://github.com/java-operator-sdk/java-operator-sdk/pull/1493


rk3rn3r:
This PR seems to cover multiple topics?

lburgazzoli:
> This PR seems to cover multiple topics?

There are some changes in the behavior of the kubernetes resource generation that I have not been able to workaround so I ended up changing also some of the kubernetes related bits.

etc/kubernetes/manifests/base/apps/cos-fleetshard-sync/observability/service-monitor.yaml
```diff
@@ -0,0 +1,56 @@
...
+spec:
+  endpoints:
+    - interval: "30s"
+      path: "/q/metrics"
+      port: "http"
+      scheme: "http"
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: "cos-fleetshard-sync"
+      app.kubernetes.io/part-of: "cos"
```
lburgazzoli, line 19: @lgarciaaco for some reason the selector for `ServiceMonitor` is not automatically updated by Quarkus anymore hence I have added a dedicate set of resources for which I have removed the matching of `app.kubernetes.io/version` and it should continue to work as before.

Let me know what you think

lgarciaaco: It looks good to me. The PodMonitors we use for the addon are in https://github.com/bf2fc6cc711aee1a0c2a/cos-tools/tree/main/observability/resources/prometheus/pod_monitors and the label_selector is the name. These are used for deployment don using the templates files generated by Quarkus, provably dev or local environment only.

lgarciaaco (approved):

oscerd (approved):

rinaldodev (approved):

PR #490, opened 2022-10-12
Mgdctrs 1379

shivamG640:
This PR has some changes where we are exposing a counter metric to count the connector state. This will help us define Availability and Error Budget SLO in the data-plane. We have exposed a new state-failed_but_ready (just for our reference and not in the ConnectorState enum) which counts datapoint for the connectors which are failing but were ready before!

### DIFF
--- a/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
+++ b/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
@@ -15,6 +15,7 @@ import org.bf2.cos.fleetshard.sync.client.FleetShardClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
@@ -24,6 +25,7 @@ public class ConnectorStatusUpdater {
     private static final Logger LOGGER = LoggerFactory.getLogger(ConnectorStatusUpdater.class);
 
     public static final String CONNECTOR_STATE = "connector.state";
+    public static final String CONNECTOR_STATE_COUNT = "connector.state.count";
 
     static final Integer CONNECTOR_STATE_READY = 1;
     static final Integer CONNECTOR_STATE_FAILED = 2;
@@ -50,7 +52,7 @@ public class ConnectorStatusUpdater {
 
             fleetManagerClient.updateConnectorStatus(connector, connectorDeploymentStatus);
 
-            LOGGER.debug("Updating Connector status metrics (Connector_id : {}, state: {})",
+            LOGGER.debug("Updating Connector status metrics (Connector_id: {}, state: {})",
                 connector.getSpec().getConnectorId(),
                 ConnectorStatusExtractor.extract(connector).getPhase());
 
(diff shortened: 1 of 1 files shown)


cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -108,5 +112,26 @@ private void measure(ManagedConnector connector, Integer connectorState) {
             .tags(tags)
             .register(registry);
 
+        Counter.builder(config.metrics().baseName() + "." + CONNECTOR_STATE_COUNT)
+            .tags(tags)
+            .tag("cos.connector.state", ConnectorStatusExtractor.extract(connector).getPhase().getValue())
+            .register(registry)
+            .increment();
+
+        if (CONNECTOR_STATE_FAILED.compareTo(connectorState) == 0) {
```
lburgazzoli, line 121: this can probably be replace by equals but maybe all the constants can be switched to a primitive type (int instead of Integer)

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -87,9 +89,11 @@ public void update(ManagedConnector connector) {
     }
 
     /*
-     * Expose a metric "cos_fleetshard_sync_connector_state" which reveals the current connector state.
+     * Expose a Gauge metric "cos_fleetshard_sync_connector_state" which reveals the current connector state.
      * Metric value of 1 implies that the connector is in Ready state. Similarly, 2 -> Failed, 3 -> Deleted,
      * 4 -> Stopped, 5 -> In Process
+     * Also exposing a Counter metrics "cos_fleetshard_sync_connector_state_count_total" which reveals each
```
lburgazzoli, line 95: where is this exposed ? is it automatically exposed from the gauge ?

lgarciaaco: micrometer appends `total` for counters I believe. `cos_fleetshard_sync_connector_state_count` in defect get exposed as `cos_fleetshard_sync_connector_state_count_total`

lgarciaaco (approved):

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -50,7 +52,7 @@ public void update(ManagedConnector connector) {
 
             fleetManagerClient.updateConnectorStatus(connector, connectorDeploymentStatus);
 
-            LOGGER.debug("Updating Connector status metrics (Connector_id : {}, state: {})",
+            LOGGER.debug("Updating Connector status metrics (Connector_id: {}, state: {})",
                 connector.getSpec().getConnectorId(),
                 ConnectorStatusExtractor.extract(connector).getPhase());
```
lburgazzoli, line 57: I think we can use the `connectorDeploymentStatus` previously computed

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorStatusUpdater.java
```diff
@@ -87,11 +89,13 @@ public void update(ManagedConnector connector) {
...
     /*
-     * Expose a metric "cos_fleetshard_sync_connector_state" which reveals the current connector state.
+     * Expose a Gauge metric "cos_fleetshard_sync_connector_state" which reveals the current connector state.
      * Metric value of 1 implies that the connector is in Ready state. Similarly, 2 -> Failed, 3 -> Deleted,
      * 4 -> Stopped, 5 -> In Process
+     * Also exposing a Counter metrics "cos_fleetshard_sync_connector_state_count_total" which reveals each
+     * state count for the connector
      */
-    private void measure(ManagedConnector connector, Integer connectorState) {
+    private void measure(ManagedConnector connector, int connectorState) {
```
lburgazzoli, line 98: Maybe also include the `ConnectorDeploymentStatus` previously computed so we can avoid an additional `ConnectorStatusExtractor.extract`

lburgazzoli (approved):

PR #506, opened 2022-10-17
MGDCTRS-1098: add basic error message processing

rinaldodev:


cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
```diff
@@ -308,7 +308,7 @@ public static Map<String, String> createSecretsData(
 
         // always enable supervising route controller, so that camel pods are not killed in case of failure
         // this way we can check it's health and report failing connectors
-        props.put("camel.main.route-controller-supervise-enabled", "true");
+        props.put("camel.main.route-controller-supervise-enabled", "false");
```
lburgazzoli, line 311: This is likely to be incorrect as it would cause the container to keep crash looping in case a component or a route cannot be started.

lburgazzoli:
@rinaldodev I move the pr to draft as we need to understand what is the root cause behind the need to disable the supervise route controller

lburgazzoli (approved):

PR #521, opened 2022-11-07
Create observability sso secret depending on environment

lgarciaaco:


rinaldodev (approved):
Not that we have strict rules around that, but second commit could use a better title.

rinaldodev (approved):

PR #522, opened 2022-11-07
MGDCTRS-1169: cos-fleetshard-operator creates an empty configmap for …

valdar:
…each ManagedConnector

this can be used to change some configurations like toggling log DEBUG level.

### DIFF
--- a/cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
+++ b/cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
@@ -28,13 +28,16 @@ import org.bf2.cos.fleetshard.operator.operand.OperandControllerMetricsWrapper;
 import org.bf2.cos.fleetshard.operator.operand.OperandResourceWatcher;
 import org.bf2.cos.fleetshard.support.exceptions.WrappedRuntimeException;
 import org.bf2.cos.fleetshard.support.metrics.MetricsRecorder;
+import org.bf2.cos.fleetshard.support.resources.ConfigMaps;
 import org.bf2.cos.fleetshard.support.resources.Resources;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.databind.JsonNode;
 
+import io.fabric8.kubernetes.api.model.ConfigMap;
 import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.OwnerReferenceBuilder;
 import io.fabric8.kubernetes.api.model.Secret;
 import io.fabric8.kubernetes.client.KubernetesClient;
@@ -133,6 +136,13 @@ public class ConnectorController implements Reconciler<ManagedConnector>, EventS
                 kubernetesClient,
                 managedConnectorOperator,
                 MetricsRecorder.of(registry, config.metrics().baseName() + ".controller.event.secrets", tags)));
+
+        eventSources.put(
+            "_configmaps",
+            new ConnectorConfigmapEventSource(
+                kubernetesClient,
+                managedConnectorOperator,
+                MetricsRecorder.of(registry, config.metrics().baseName() + ".controller.event.configmaps", tags)));
 
         eventSources.put(
             "_operators",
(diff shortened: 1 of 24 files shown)


cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
```diff
@@ -393,6 +403,50 @@ private UpdateControl<ManagedConnector> handleAugmentation(ManagedConnector conn
...
+            // copy operator type
+            Resources.setLabel(
+                secret,
+                LABEL_OPERATOR_TYPE,
+                connector.getMetadata().getLabels().get(LABEL_OPERATOR_TYPE));
+            try {
+                this.kubernetesClient.configMaps()
+                    .inNamespace(configMap.getMetadata().getNamespace())
+                    .withName(configMap.getMetadata().getName())
+                    .createOrReplace(configMap);
```
lburgazzoli, line 443: This should probably be just `create` instead of `createOrReplace` as it may override an existing configmap in case it gets created between the check for presence and the invocation of `createOrReplace`

valdar: done

lburgazzoli (approved):

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
```diff
@@ -393,6 +403,50 @@ private UpdateControl<ManagedConnector> handleAugmentation(ManagedConnector conn
...
+                LABEL_CONNECTOR_ID, connector.getSpec().getConnectorId(),
+                LABEL_DEPLOYMENT_ID, connector.getSpec().getDeploymentId());
+
+            Resources.setOwnerReferences(
+                configMap,
+                connector);
+
+            // copy operator type
+            Resources.setLabel(
+                secret,
```
rinaldodev, line 436: Is this really supposed to be the `secret`, or should it be the `configMap`?
If it is the `configMap`, can it be done in L424? Also, I'd also suggest moving the secret query in L401 to after this `if` block.

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
```diff
@@ -393,6 +403,50 @@ private UpdateControl<ManagedConnector> handleAugmentation(ManagedConnector conn
...
+                LABEL_OPERATOR_TYPE,
+                connector.getMetadata().getLabels().get(LABEL_OPERATOR_TYPE));
+            try {
+                this.kubernetesClient.configMaps()
+                    .inNamespace(configMap.getMetadata().getNamespace())
+                    .withName(configMap.getMetadata().getName())
+                    .create(configMap);
+            } catch (Exception e) {
+                LOGGER.warn("", e);
```
rinaldodev, line 445: Why the use of log-and-rethrow here? Since we don't add anything to the message, can't we just not catch?

rinaldodev (changes requested):

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
valdar: I was just being consistent with https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/e13e6f5bd656c778a6b718f8d09100f1f2b60c82/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ConnectorDeploymentProvisioner.java#L333-L336 but you have a point.

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
valdar: Ah! Good catch. This also highlights that some tests are missing.

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
rinaldodev: I also don't know why we have this one :P

lburgazzoli: AFAIK it was to avoid the exception to be swallowed, maybe it is not more the case but it was in the past

valdar: done

cos-fleetshard-operator/src/main/java/org/bf2/cos/fleetshard/operator/connector/ConnectorController.java
valdar: done

rinaldodev (approved):

valdar:
@rinaldodev since you made me think about it, I preferred to swap secret and configmap order in the controller so that the configmap is not created if there are problems with the secret.

valdar:
added configmaps permissions to the cos-fleetshard-camel serviceaccount

oscerd (approved):

rinaldodev (approved):

PR #523, opened 2022-11-08
Removed unecessary finalizer in observability resource creation.

valdar:


valdar:
Usually adding finalizers is responsibility of the operator that manage it. This seems to confirm it https://github.com/redhat-developer/observability-operator/blob/42d5675fe990d85849f3f2ed46b9bbffa294f167/controllers/observability_controller.go#L91-L96

rinaldodev (approved):

PR #530, opened 2022-11-14
MGDCTRS-1671 Failing connector should not stuck in provision phase when connection to Kafka Cluster is not possible

rk3rn3r:
relates to https://issues.redhat.com/browse/MGDCTRS-1671


rk3rn3r:
Build error is unrelated. LGTM

PR #535, opened 2022-11-23
MGDCTRS-1599: configure camel connector logs based on ConfigMap

rinaldodev:


cos-fleetshard-api/model/src/main/java/org/bf2/cos/fleetshard/api/DeploymentSpec.java
```diff
@@ -29,6 +30,7 @@ public class DeploymentSpec {
     private SchemaRegistrySpec schemaRegistry;
     private String desiredState;
     private String secret;
+    private String configMapChecksum;
```
lburgazzoli, line 33: What is the purpose to store the configmap checksum in the deployment spec ?

cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
```diff
@@ -340,6 +335,40 @@ public static Map<String, String> createSecretsData(
                 cfg.exchangePooling().exchangeFactoryStatisticsEnabled());
         }
 
+        // configure the empty config map created for logging
+        final ConfigMap configMap = connectorConfiguration.getConfigMap();
```
lburgazzoli, line 339: if I properly understand this block, we take data from the configmap and copy it over the secret that already exists but another option would be to mount the configmap and let the runtime load any property that is defined (so no need to translate `log.level` into specific quarkus properties (which, as side effect, would allow to also add specific properties to tweak the behavior  of the connector if needed)

cos-fleetshard-api/model/src/main/java/org/bf2/cos/fleetshard/api/DeploymentSpec.java
rinaldodev: To trigger reify when it's contents change, same strategy we use for the integration secret. If you have a better alternative let me know.

cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
rinaldodev: I recall part of this being discussed and one of the points was how much control we want to have over it. If we do that, we lose control entirely.
Also, if we do go down that route, I'd question many of the current exposed/configured properties in connectors or even operators, since we could just put everything in the configmaps and expose all of it.

rinaldodev: @valdar wdyt?

cos-fleetshard-api/model/src/main/java/org/bf2/cos/fleetshard/api/DeploymentSpec.java
lburgazzoli: This should happen anyway as the configmap is among the event sources so every change to the configmap should trigger a reconcile loop but I think the issue here is that we don't go through the augmentation phase ?

cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
lburgazzoli: > I recall part of this being discussed and one of the points was how much control we want to have over it. If we do that, we lose control entirely. Also, if we do go down that route, I'd question many of the current exposed/configured properties in connectors or even operators, since we could just put everything in the configmaps and expose all of it.

Ok so the implemented behavior is fine with me.

The question was mainly related to how the same config map approach works for the sync & operators where we don't have any kind of control whereas here we decided to go into a more controlled way.

cos-fleetshard-api/model/src/main/java/org/bf2/cos/fleetshard/api/DeploymentSpec.java
rinaldodev: Reconciliation is triggered but not reify, as we only trigger reify when there are changes specifically to the deployment spec. We could change that behavior as well if you see fit.

lburgazzoli: yeah, that something we must change at some point but not in this pr

lburgazzoli (approved):

cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
valdar: @lburgazzoli @rinaldodev I think that the idea behind the 2 different approaches was based on how often we foreseen to use them. The idea is that connectors are the more  likely to be placed in debug mode and so the one for which we want to prevent erroneous configuration the most.

That said, the counter argument is that something is not working if we need to activate debug mode, so as long as the correct values are documented in SOPs we should be good and gain flexibility for future unforeseen needs. Also there is the argument of consistency across system components...

valdar: > Also, if we do go down that route, I'd question many of the current exposed/configured properties in connectors or even operators, since we could just put everything in the configmaps and expose all of it.

I am not sure I completely get you: the fact that there is the possibility of overriding properties putting them in a configmap dose not make the current way of using some properties less effective, I think.

rinaldodev: @valdar it doesn't make them less effective, I'd just opt for a single one in favor of simplicity.

I understand we can then move forward with the current solution.

rinaldodev: @valdar and I have discussed this and opted for a more flexible solution.

rinaldodev:
Converting to draft as it lacks a README update.

rinaldodev:
@valdar please take a look when you have the time

rinaldodev:
@lburgazzoli ConfigMap is now flexible and can be used to add any property to the connector.

lburgazzoli (approved):

valdar (approved):

Issue #536, opened 2022-11-24
Observatorium config secret is created with wrong name


lgarciaaco:
Fix typo in https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/e25352ceb73816727c9d0dab967f906fed12314d/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/client/FleetShardObservabilityClient.java#L84

It should be `getName` not namespaces.

When the observability secret is present the sync is unable to start with:
```
2022-11-24 11:27:38,803 INFO  [org.bf2.cos.fleetshard.sync.client.FleetShardObservabilityClient] (main) Configuring Observatorium SSO secret                                                          
2022-11-24 11:27:39,212 ERROR [io.quarkus.runtime.Application] (main) Failed to start application (with profile prod): io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://172.30.0.1/api/v1/namespaces/redhat-openshift-connectors/secrets/redhat-openshift-connectors. Message: secrets "redhat-openshift-connectors" not found. Received status: Status(apiVersion
=v1, code=404, details=StatusDetails(causes=[], group=null, kind=secrets, name=redhat-openshift-connectors, retryAfterSeconds=null, uid=null, additionalProperties={}), kind=Status, message=secrets "
redhat-openshift-connectors" not found, metadata=ListMeta(_continue=null, remainingItemCount=null, resourceVersion=null, selfLink=null, additionalProperties={}), reason=NotFound, status=Failure, additionalProperties={})
```

PR #509, opened 2022-10-19
MGDCTRS-1040 Strimzi upgrade 0.32.0

rk3rn3r:
MGDCTRS-1040 Strimzi upgrade to 0.32.0. Upgrading Java dependencies and Strimzi operator version

relates to #468
closes https://issues.redhat.com/browse/MGDCTRS-1040

PR #537, opened 2022-11-24
Fix typo preventing sync to start when cos.observability.environment secret is found

lgarciaaco:
Interestingly the tests didn't fail and in dev the problem is not present 🤔 

Fixes https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/issues/536


lburgazzoli (approved):

rinaldodev:
We should consider upgrading the fabric8 client to the version that allows us to use the resource(secret) method to avois those things. It's too error prone. @lburgazzoli do you know what version that is?

rinaldodev (approved):

lburgazzoli:
> We should consider upgrading the fabric8 client to the version that allows us to use the resource(secret) method to avois those things. It's too error prone. @lburgazzoli do you know what version that is?

Things will be much better in 6.x but it requires the quarkus-operator-sdk 5.x which is currently in beta.

There is a [PR](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/pull/509) with a combination of dependencies that seems to work but I would be much more comfortable waiting for 5.x

PR #542, opened 2022-11-29
MGDCTRS-688: Suport for leader election

lburgazzoli:
This PR aims to add support for leader election mechanism in the fleetshard components by using the [lease/v1 APIs](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/lease-v1/) to prevent multiple operator or sync to reconcile same resource which would lead to undefined behaviors. This would also allow to have multiple replicas of any data plane component in case it is needed.

As side effect this PR also introduce some e2e tests where the operators are installed on a Kind cluster which helps detecting Kubernetes resource misconfigurations


lburgazzoli:
Not needed at this stage

PR #557, opened 2022-12-19
MGDCTRS-1600: make possible to enable debug for dbz connectors

rinaldodev:


rinaldodev:
@rk3rn3r

valdar (approved):

pom.xml
```diff
@@ -194,6 +195,11 @@
                 <artifactId>commons-text</artifactId>
                 <version>${commons-text.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
```
rk3rn3r, line 200: Out of curiosity: What is this used for?

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
```diff
@@ -2,12 +2,7 @@
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.TreeMap;
+import java.util.*;
```
rk3rn3r, line 5: Are we okay with `*` imports?

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
```diff
@@ -189,6 +178,9 @@ protected List<HasMetadata> doReify(
                 .addToRequests("memory", new Quantity("256Mi"))
                 .addToLimits("cpu", new Quantity("500m"))
                 .addToLimits("memory", new Quantity("1Gi"))
+                .build())
+            .withLogging(new InlineLoggingBuilder()
+                .withLoggers(loggingMap)
```
rk3rn3r, line 183: This seems doable. Would it be possible to use external logging configuration via a ConfigMap and update the ConfigMap in case of an enabled DEBUG mode?
We don't have to do this now, but maybe we can explore this solution in the future? Could you create a ticket for this?

pom.xml
rinaldodev: `StringUtils.normalizeSpace` when logging.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
rinaldodev: Changed back to the individual imports. 

This was done by IntelliJ. My IDE is configured using [this](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/main/cos-build-tools/src/main/resources/eclipse-format.xml) file, but apparently it doesn't include the [import order](https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/main/cos-build-tools/src/main/resources/eclipse.importorder).

I have now installed [this plugin](https://github.com/krasa/EclipseCodeFormatter#instructions) in IntelliJ to be able to use the .importorder as well, but even then it generates different imports from the current ones. I'd suggest we find a better way to auto-organize those so it is never anymore a source of discussion.

cos-fleetshard-operator-debezium/src/main/java/org/bf2/cos/fleetshard/operator/debezium/DebeziumOperandController.java
rinaldodev: Issue MGDCTRS-1829

rk3rn3r (approved):
lgtm

PR #567, opened 2023-01-12
chore(it): update referenced kameletbinding crd

lburgazzoli:


rinaldodev (approved):
Do you recall why we needed to add that to the IT overlay?

lburgazzoli:
> Do you recall why we needed to add that to the IT overlay?

it is because we only need one CRD for out integration tests, I guess we can reference the CRDs from the base layer.
Will explore that option later if fine with you

rinaldodev:
> > Do you recall why we needed to add that to the IT overlay?
> 
> it is because we only need one CRD for out integration tests, I guess we can reference the CRDs from the base layer. Will explore that option later if fine with you

Ah, yes. Ok, we definitely should do that.

rinaldodev (approved):

PR #578, opened 2023-01-16
feat: add basic e2e test

lburgazzoli:


valdar (approved):

lgarciaaco (approved):

etc/kubernetes/manifests/overlays/e2e/kustomization.yaml
```diff
@@ -0,0 +1,86 @@
...
+    apiVersion: camel.apache.org/v1
+    kind: IntegrationPlatform
+    metadata:
+      name: camel-k
+
+patches:
+  #
+  # sync
+  #
+  - target:
```
rinaldodev, line 19: Can't we change the target to `kind: Deployment` only and use the same for all of them?

rinaldodev (approved):

etc/kubernetes/manifests/overlays/e2e/kustomization.yaml
lburgazzoli: I think it was not possible last time I checked, I may explore other transformers later on

PR #584, opened 2023-01-20
MGDCTRS-1113: Added Metrics housekeeping

shivamG640:
Added metrics housekeeper job to remove deleted connectors metrics once a day. Note that it only removes the metrics that are 6 hours older.

### DIFF
--- a/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/FleetShardSyncConfig.java
+++ b/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/FleetShardSyncConfig.java
@@ -157,6 +157,15 @@ public interface FleetShardSyncConfig {
         @WithDefault("30s")
         @WithConverter(DurationConverter.class)
         Duration housekeeperInterval();
+
+        /**
+         * Determine how often the cleaning of old metrics should happen.
+         *
+         * @return the metricshousekeeping interval
+         */
+        @WithDefault("24h")
+        @WithConverter(DurationConverter.class)
+        Duration metricsHousekeeperInterval();
     }
 
     interface Addon {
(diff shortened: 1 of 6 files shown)


rinaldodev:
I don't belive there is a need to implement a new housekeeper, just a new Service. The current housekeeper will start any class extending the Service interface: https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/blob/main/cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/housekeeping/Housekeeper.java#L45-L49

shivamG640:
@lburgazzoli, Rinaldo, and I discussed the previous comment and waiting for your comment on it!
The purpose of the new housekeeper is to clean the metric registry by just running once a day, unlike the existing scheduler which keeps running every 30 seconds. There is a query to fetch the deleted connectors metrics and deletion of metrics only happens if the metrics are 6 hours older. So What I feel is using the existing scheduler for this purpose might add an extra unnecessary load of querying the deleted connectors every 30 secs and that can be a problem depending upon the number of connectors getting deleted from the data plane.

heiko-braun:
@lburgazzoli Mind a review?

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/housekeeping/MetricsHousekeeperJob.java
```diff
@@ -0,0 +1,76 @@
...
+                            }
+                        }
+
+                        // removing the gauge type metrics
+                        registry.remove(gauge);
+                        LOGGER.info("Deleted all connector metrics: {}", id);
+                    }
+                }
+            }
+        } catch (Exception ex) {
```
lburgazzoli, line 71: We should probably re-throw a `JobExecutionException` with an appropriate re-scheduler otherwise, the cleanup will be repeated after 24h

cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/FleetShardSyncConfig.java
```diff
@@ -157,6 +157,15 @@ interface Resources {
...
         Duration housekeeperInterval();
+
+        /**
+         * Determine how often the cleaning of old metrics should happen.
+         *
+         * @return the metricshousekeeping interval
+         */
+        @WithDefault("24h")
+        @WithConverter(DurationConverter.class)
+        Duration metricsHousekeeperInterval();
```
lburgazzoli, line 168: just as a note: at some point, we may think about adding a generic `daily` housekeeping runner

lburgazzoli:
Added two small comments that can be addressed on a subsequent PR
In addition it would be nice to have a test for this feature.

PR #595, opened 2023-02-06
chore: fix deprecation warnings

lburgazzoli:


oscerd (approved):

cos-fleetshard-it/cucumber/src/main/java/org/bf2/cos/fleetshard/it/cucumber/ConnectorSteps.java
```diff
@@ -573,9 +573,9 @@ public void deploy_secret_with_uow(String uow, Map<String, String> content) {
             uow);
 
         ctx.secret(
-            kubernetesClient.resources(Secret.class)
+            kubernetesClient.resource(ctx.secret())
                 .inNamespace(ctx.connectorsNamespace())
```
rinaldodev, line 577: Do you know how the new API handles namespace name in metadata vs using this method?

rinaldodev (approved):

cos-fleetshard-it/cucumber/src/main/java/org/bf2/cos/fleetshard/it/cucumber/ConnectorSteps.java
lburgazzoli: The API should use the namespace that is given as part of the resource, the additional `inNamespace` should force a namespace regardless of what is defined by the resource. 

I may clean up ting even further later on but we must check if all the resources have a namespace defined.

lburgazzoli: Got some more info:

```
the logic should match the kubectl logic.  That is if you specify a -n (inNamespace) that will be used, if not it will use the namespace of the item.  If the item has no namespace, then it will use the default namespace of the context.
```

rinaldodev: Thanks.

PR #604, opened 2023-02-14
MGDCTRS-2016, 1896

shivamG640:


shivamG640:
This PR contains the fix for the [issue](https://issues.redhat.com/browse/MGDCTRS-2016) and tests for the metrics housekeeper feature.

rinaldodev (approved):
About the test: I think an alternative would be letting the sync create the Managed Connector CR by stubbing the call to the fleetmanager's deployments endpoint and make it return a new deployment. But this version LGTM as well.

shivamG640:
Thanks, @rinaldodev . I am merging this PR so that this gets deployed with today's stage deployment!

PR #612, opened 2023-03-06
chore: include migrate field

lgarciaaco:
Version 4 of the Observability Operator [introduces the field `Migrated` in the CRD Status block](https://github.com/redhat-developer/observability-operator/blob/main/api/v1/observability_types.go#L128) and this needs to be reflected on the sync


lgarciaaco:
Fixes https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/issues/613

lburgazzoli:
@lgarciaaco is this backward compatible ?

lgarciaaco:
> @lgarciaaco is this backward compatible ? if it matters of course

it doesn't matter

lburgazzoli (approved):

PR #636, opened 2023-03-29
First draft of Processors Provisioning mostly duplicated code from Connector, compiling

lucamolteni:


cos-fleetshard-sync/src/main/java/org/bf2/cos/fleetshard/sync/resources/ProcessorStatusUpdater.java
```diff
@@ -0,0 +1,57 @@
...
+    FleetShardClient connectorClient;
+    @Inject
+    FleetShardSyncConfig config;
+
+    public void update(ManagedProcessor processor) {
+        LOGGER.debug("Update processor status (name: {}, phase: {})",
+            processor.getMetadata().getName(),
+            processor.getStatus().getPhase());
+
+        fleetManagerClient.getProcessorDeployments(Long.parseLong(processor.getSpec().getDeploymentId()),
```
lburgazzoli, line 31: I'm not very sure I get this right, I guess it is only a placeholder at this stage ?

cos-fleetshard-sync/src/main/resources/application.properties
```diff
@@ -1,5 +1,5 @@
 quarkus.banner.enabled = false
-quarkus.log.level = INFO
+quarkus.log.level = DEBUG
```
lburgazzoli, line 2: Remember to put it back to INFO before merge

PR #637, opened 2023-03-30
MGDCTRS-516 : health check and error handler not working when kafka instance is del…

vaibhavjainwiz:
…eted

Jira : https://issues.redhat.com/browse/MGDCTRS-516

### DIFF
--- a/cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
+++ b/cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
@@ -373,25 +373,20 @@ public final class CamelOperandSupport {
     }
 
     public static void computeStatus(ConnectorStatusSpec statusSpec, KameletBindingStatus kameletBindingStatus) {
-        if (kameletBindingStatus.phase != null) {
-            switch (kameletBindingStatus.phase.toLowerCase(Locale.US)) {
-                case KameletBindingStatus.PHASE_READY:
-                    statusSpec.setPhase(ManagedConnector.STATE_PROVISIONING);
-                    if (kameletBindingStatus.conditions != null) {
-                        boolean readyCondition = kameletBindingStatus.conditions.stream()
-                            .anyMatch(cond -> "Ready".equals(cond.getType())
-                                && "True".equals(cond.getStatus()));
-                        if (readyCondition) {
-                            statusSpec.setPhase(ManagedConnector.STATE_READY);
-                        }
+        statusSpec.setPhase(ManagedConnector.STATE_PROVISIONING);
+        if (kameletBindingStatus.conditions != null) {
+            Optional<Condition> readyConditionOpt = kameletBindingStatus.conditions.stream()
+                .filter(cond -> "Ready".equals(cond.getType())).findAny();
+            if (readyConditionOpt.isPresent()) {
+                Condition readyCondition = readyConditionOpt.get();
+                boolean klbReady = "True".equals(readyCondition.getStatus());
+                if (klbReady) {
+                    statusSpec.setPhase(ManagedConnector.STATE_READY);
+                } else {
+                    if (isErrorReasons(readyCondition.getReason())) {
+                        statusSpec.setPhase(ManagedConnector.STATE_FAILED);
                     }
-                    break;
-                case KameletBindingStatus.PHASE_ERROR:
-                    statusSpec.setPhase(ManagedConnector.STATE_FAILED);
-                    break;
-                default:
-                    statusSpec.setPhase(ManagedConnector.STATE_PROVISIONING);
-                    break;
+                }
             }
         }
 
@@ -414,6 +409,13 @@ public final class CamelOperandSupport {
                     "False".equals(cond.getStatus()))
             .findFirst()
             .ifPresent(ConditionMessageImprover::improve);
+    }
+
+    private static boolean isErrorReasons(String reason) {
+        if (reason == null) {
+            return false;
+        }
+        return List.of("Error", "RuntimeNotReady").contains(reason);
     }
 
     public static ObjectNode createErrorHandler(
--- a/cos-fleetshard-operator-camel-it/src/test/java/org/bf2/cos/fleetshard/operator/it/camel/glues/KameletBindingSteps.java
+++ b/cos-fleetshard-operator-camel-it/src/test/java/org/bf2/cos/fleetshard/operator/it/camel/glues/KameletBindingSteps.java
@@ -40,8 +40,8 @@ public class KameletBindingSteps extends StepsSupport {
     }
 
     @SuppressWarnings("unchecked")
-    @When("the klb phase is {string} with conditions:")
-    public void klb_phase_and_conditions(String phase, DataTable table) {
+    @When("the klb with conditions:")
+    public void klb_phase_and_conditions(DataTable table) {
 
         // TODO: investigate using KubernetesClient.resources(KameletBinding.class) result in a bad patch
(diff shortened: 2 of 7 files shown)


cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java
```diff
@@ -373,25 +373,19 @@ public static Optional<KameletBinding> lookupBinding(KubernetesClient client, Ma
...
+        statusSpec.setPhase(ManagedConnector.STATE_PROVISIONING);
+        if (kameletBindingStatus.conditions != null) {
+            Optional<Condition> readyConditionOpt = kameletBindingStatus.conditions.stream().filter(cond -> "Ready".equals(cond.getType())).findAny();
+            if (readyConditionOpt.isPresent()) {
+                Condition readyCondition = readyConditionOpt.get();
+                boolean klbReady =  "True".equals(readyCondition.getStatus());
+                if (klbReady) {
+                    statusSpec.setPhase(ManagedConnector.STATE_READY);
+                } else {
+                    if(!readyCondition.getMessage().isEmpty()) {
```
lburgazzoli, line 385: I'm not sure if this is 100% correct as I think - but it needs to be double checked - the ready condition may be `False` but because the deployment is in progress hence the only presence of a message is not an indication that there is a failure. I believe we must check the content of the message at this stage.

But @rinaldodev may know better

vaibhavjainwiz: With failure `KameletBindingStatus` looks like
```
status:
  conditions:
    - lastTransitionTime: '2023-03-30T09:33:12Z'
      lastUpdateTime: '2023-03-30T09:33:12Z'
      message: >-
        [Pod mctr-cgila9m8qbfm7ltin430-9f686bb74-2mml6 runtime is not ready:
        map[bootstrap.servers:vajain-rho-cgijrtpm-iq--tu-nbaa.bf2.kafka.rhcloud.com:443
        check.kind:READINESS client.id:consumer-cgila9m8qbfm7ltin430-1
        group.id:cgila9m8qbfm7ltin430 route.id:cgila9m8qbfm7ltin430-source
        topic:vajain_topic]]
      reason: RuntimeNotReady
      status: 'False'
      type: Ready
  observedGeneration: 1
  phase: Ready
  replicas: 1
  selector: camel.apache.org/integration=mctr-cgila9m8qbfm7ltin430
  ```

Comment (vaibhavjainwiz) [reply cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java:385]: I had done a quick test with the changes of this PR, its working fine for atleast this particular scenario. 
But I agree with @lburgazzoli concern, might be its not valid for all scenario. Best I could think is to run the all intgeration test and see if everything works as expected.

Comment (rinaldodev) [reply cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java]: I agree with Luca here. We need to make sure that having a message actually means a failure.

By looking at Camel K code, it looks we could do that if we filter out those 2 phases:

```
	// KameletBindingPhaseNone --
	KameletBindingPhaseNone KameletBindingPhase = ""
	// KameletBindingPhaseCreating --
	KameletBindingPhaseCreating KameletBindingPhase = "Creating"
```

During the initialization, when the KameletBinding is in one of those 2 phases, it will be "Ready=false". If at that same moment a message is present, that may not represent a failure, just a deployment that is in progress.

https://github.com/bf2fc6cc711aee1a0c2a/cos-fleetshard/pull/637/files#diff-8c06870edc5b0b037fa5742434a6a98dd9fd3fe3b90df639f9847382c3082239R386-R388

Comment (vaibhavjainwiz) [reply cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java]: @lucamolteni @rinaldodev 
While analysing Camel-K operator code, I notice that Camel-k has already fixed this issue in `v1.12.0` to mark Integration PHASE to `ERROR` if POD is not recoverable.
I also tested our application with Camel-K `v1.12.0`, this behaviour is wroking as expected and `cos-fleetshard-camelk-operator` has marked the `ManagedConnector` resource to `failed` state without any change of our code.
https://github.com/apache/camel-k/blob/main/pkg/controller/integration/monitor.go#L423

Could we upgrade Camel-K operator version to `v1.12.0`? or we still need to implement some hotfix in our application?

Comment (vaibhavjainwiz) [reply cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java]: I had push the changes to monitor Error message `runtime not ready`.

Comment (vaibhavjainwiz) [reply cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java]: I made some more changes in my PR
rather then depends on `message`, I put the check on `reason`

`message` values are derived from different objects such as pod Condition, Container Condition, Knative Condition, Integration Kit Condition etc. Its getting very hard to track what exact message each of these object sends
I found that Camel-K already have some logic to segregate these failure condition and then always set specific `Reason = Error` when situation is un-recoverable

**Comment by vaibhavjainwiz** [comment]
@lburgazzoli Could you please review the PR

File: cos-fleetshard-operator-camel/src/main/java/org/bf2/cos/fleetshard/operator/camel/CamelOperandSupport.java [hunk 2]
```diff
@@ -373,25 +373,20 @@ public static Optional<KameletBinding> lookupBinding(KubernetesClient client, Ma
...
+                boolean klbReady = "True".equals(readyCondition.getStatus());
+                if (klbReady) {
+                    statusSpec.setPhase(ManagedConnector.STATE_READY);
+                } else {
+                    if (isErrorReasons(readyCondition.getReason())) {
+                        statusSpec.setPhase(ManagedConnector.STATE_FAILED);
                     }
-                    break;
-                case KameletBindingStatus.PHASE_ERROR:
-                    statusSpec.setPhase(ManagedConnector.STATE_FAILED);
```
lburgazzoli, line 390: isn't this still a thing ? i.e if there is a technical error while reconciling the integration

@squakez what do you think ?

vaibhavjainwiz: As discussed, I completely remove dependency from KBL PHASE and move all checks on conditions only. I am hoping if there is some technical error then KBL Ready condition would be set to `false`.

lburgazzoli: Yep but I'm not very sure that the ready condition is alway present, as example in the kamelet binding initialization reconcile action from camel-k (https://github.com/apache/camel-k/blob/d5b9466248fdcc2eedf6489cec978c6b527869ff/pkg/controller/kameletbinding/initialize.go#L52), in case of an issue creating the integration, the condition that is set is `IntegrationError` and I don't see any track of a `Ready` condition.

btw, I thin the type of thjat condition is very misleading as then you gonna have `Type: IntegrationError` and `Status: False`

squakez: Yes, also consider that it's the canonical way we rebuild the KameletBindings (ie, wiping it off the status completely), though I am not sure if that is something that may affect you.

rinaldodev (approved):

Srihari1192 (approved):

PR #641, opened 2023-04-04
[MGDCTRS-2210] - Implement ManagedProcessor CRD

vaibhavjainwiz:
https://issues.redhat.com/browse/MGDCTRS-2210


lburgazzoli:
This is pretty much the same as a `ManagedConnector` so it is pretty ok, however in the `ManagedConnector` CRD there are some tech depth that it may be better not to repeat:

- the spec of the connectors is blindly copied in a secrtet, this has been done in order oto avoid leaking sensitive information in the CR, however for processors, we expect the route definition not to include any sensitive info hence, the definition and the shard meta can be embedded in the CR which greatly simplify things
- there is `*DeploymentSpec` that is duplicated in the Spec and Status section of the CRD and for connectors, it was done to help detecting drifts, however that "drift detection" is far from be ideal and I would really recommend to get rid of it and implement a continuous reconciliation via patch

vaibhavjainwiz:
Thanks for headsup @lburgazzoli 
> the spec of the connectors is blindly copied in a secrtet, this has been done in order oto avoid leaking sensitive information in the CR, however for processors, we expect the route definition not to include any sensitive info hence, the definition and the shard meta can be embedded in the CR which greatly simplify things

In our secret we are planning to keep `client_id` and `client_secret`

 > there is *DeploymentSpec that is duplicated in the Spec and Status section of the CRD and for connectors, it was done to help detecting drifts, however that "drift detection" is far from be ideal and I would really recommend to get rid of it and implement a continuous reconciliation via patch
 
 I had removed the `DeploymentSpec` from status. We would use the patch in implementation.

lburgazzoli:
> Thanks for headsup @lburgazzoli
> 
> > the spec of the connectors is blindly copied in a secrtet, this has been done in order oto avoid leaking sensitive information in the CR, however for processors, we expect the route definition not to include any sensitive info hence, the definition and the shard meta can be embedded in the CR which greatly simplify things
> 
> In our secret we are planning to keep `client_id` and `client_secret` only. [gist.github.com/vaibhavjainwiz/2c8d15e19e1f7dde8a9a73a8ef090cb1#file-managedprocessor-yaml-L49](https://gist.github.com/vaibhavjainwiz/2c8d15e19e1f7dde8a9a73a8ef090cb1#file-managedprocessor-yaml-L49)
> 

You can even include the route inside the spec, it make things easier to detect changes

> > there is *DeploymentSpec that is duplicated in the Spec and Status section of the CRD and for connectors, it was done to help detecting drifts, however that "drift detection" is far from be ideal and I would really recommend to get rid of it and implement a continuous reconciliation via patch
> 
> I had removed the `DeploymentSpec` from status. We would use the patch in implementation.

You can remove also from spec and flatten the fields

vaibhavjainwiz:
> You can remove also from spec and flatten the fields
@lburgazzoli Does it make sense if I also remove `DeploymentId` & `DeploymentResourceVersion` as well?

lburgazzoli:
I don't know, that's depends if you plan to use them or not, In RHOC the resources version is used to optimize polling from the fleet manager

lucamolteni:
> I don't know, that's depends if you plan to use them or not, In RHOC the resources version is used to optimize polling from the fleet manager

Polling in the fleet manager is done by the sync, not the operator so we might assume that the operator could avoid using deployments entirely

lburgazzoli:
Yeah, I know that but for connectors, the sync uses the ManagedConnector CRs to compute what it the lates resource version known by the data plane.

Do you have plans to use other machinery ?

lucamolteni:
> Yeah, I know that but for connectors, the sync uses the ManagedConnector CRs to compute what it the lates resource version known by the data plane.
> 
> Do you have plans to use other machinery ?

The simplest solutions that comes to my mind is to copy just the resource version than the whole deployment. This decouples the operator from the concept of deployments and simplify the boundaries a little bit. But I suggest to start working on the actual operator impl to see if this works or not, rather than discussing at the data level

vaibhavjainwiz:
Currently I am adding both `DeploymentId` and `DeploymentResourceVersion` to PR. We could optimize it later.

PR #647, opened 2023-04-13
MGDCTRS-1747: Enabling CRs filtering for Stimzi Operator

shivamG640:
STRIMZI_CUSTOM_RESOURCE_SELECTOR is the env var used to enable the Strimzi operator to filter CRs with a specific label in it. And found out that "cos.bf2.org/connector.operator" is the appropriate label for the filtering purpose.

Also disabling the leader election mechanism for the operator as the other operator(2nd strimzi operator) when deployed for the migration of the connectors, elects the first one as the leader irrespective of the different env var in it and acts as a backup resulting in the CRs not getting reconciled when updated with the new operator-id for reconciliation.

### DIFF
--- a/etc/kubernetes/manifests/base/apps/strimzi/operator/kustomization.yaml
+++ b/etc/kubernetes/manifests/base/apps/strimzi/operator/kustomization.yaml
@@ -4,7 +4,7 @@ generatorOptions:
 
 commonLabels:
   app: strimzi
- 
+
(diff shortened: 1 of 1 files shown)


rinaldodev (approved):

rk3rn3r (approved):
LGTM

valdar (approved):

PR #649, opened 2023-04-18
MGDCTRS-2282: remove observability logic from the sync

lgarciaaco:
Since we are installed together with RHOSAK now, they bring their own Observability CR and there can only be one per cluster. This PR removes all resources related with the Observability stack.


rinaldodev (approved):
LGTM. Can you wait until we stabilize the current release before merging this?

